PR #2, opened 2019-02-27
Bug 1529631 & Bug 1527962

markcor:
Add support for Mozilla Build. Includes:
Package installation 
Hg upgrade and configuration 
Set priority for Hg and Python via registry 
Set caching location for included packages
Granting symlink access bassed off of included components. Needed for use of other components.  
Supporting files

Add support for Mozilla Maintenance service. Includes:
Installation of Service 
Installation and registry entries of needed certs 
Opening permissions for the Mozilla maintenance  registry key. 

Also add the unless parameter to win_packages::win_exe_pkg. It may be useful in the future.

And removing double slashes from custom facts for directory paths. They were serving no purpose.


markcor:
I am going to remove the unless parameter in win_packages::win_exe_pkg in the next pull request. It throws warnings.

dividehex:
@markcor could you also add a README.md to the moz maintence service module pointing at https://support.mozilla.org/en-US/kb/what-mozilla-maintenance-service .

PR #3, opened 2019-03-07
Bootstrap

dividehex:


dividehex:
@markcor Mark, please review the Hiera portion of the PR.  I've moved your windows specific hiera data to a windows specific yaml file.

@davehouse Dave, please review the shell script

provisioners/bootstrap_posix.sh
```diff
@@ -0,0 +1,189 @@
...
+            # touch log file to see if we have write access and fail otherwise
+            touch "$LOG_PATH" || fail "Can't write log to ${LOG_PATH}"
+            exec >"$LOG_PATH" 2>&1
+            ;;
+        :)
+            echo "Option -$OPTARG requires an argument" >&2
+            exit 1
+            ;;
+    esac
+done
```
davehouse: Nice. This script is clean and clear; Much better than how I'd write it at first pass -- I would not have provided the help or done any pretty opt handling (screw the user)!

provisioners/bootstrap_posix.sh
```diff
@@ -0,0 +1,197 @@
...
+    TMP_DL_DIR=$(mktemp -d -t puppet_download)
+    [ -d "${TMP_DL_DIR}" ] || fail "Failed to mktemp download dir"
+
+    # Download the puppet repo tarball directly from github
+    # We don't use git because some oses don't have git installed by default
+    # In the future, we may publish master branch to s3 or some other highly available service since
+    # Github has rate limits on downloads.
+    while true; do
+        echo "Downloading puppet repo: ${PUPPET_REPO}"
+        if HTTP_RES_CODE=$(curl -sL $PUPPET_REPO -o "${TMP_DL_DIR}/puppet.tar.gz" -w "%{http_code}") && [[ $HTTP_RES_CODE = "200" ]]; then
```
davehouse: Could you use curl's retry and back-off instead or in addition to this?

data/os/Windows.yaml
```diff
@@ -0,0 +1,18 @@
...
+# log aggregators
+win_dc_log_aggregator: "log-aggregator.srv.releng.%{facts.custom_win_mozspace}.mozilla.com"
+
+# ntp servers
+win_datacenterntp: "infoblox1.private.%{facts.custom_win_mozspace}.mozilla.com"
+
+# package managment
+win_ext_pkg_src: "https://s3.amazonaws.com/windows-opencloudconfig-packages/RoninPackages"
+
+# Site source url
```
markcor: Let's drop anything pointed to markcor/ronin-win-base.

manifests/site.pp
```diff
@@ -0,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Default to 0:0, 0644 on POSIX, and root, 0644 on Windows
+case $::operatingsystem {
+    'Windows': {
```
markcor: Let's not set default ownership on Windows. This should be inheritance from the parent directory by default, or set for specific user in the related code. Also note there is no longer root account on any Windows platforms.

provisioners/bootstrap_posix.sh
dividehex: I could but at this time I want the logic handled in the shell script itself.  By having it in the shell script, I don't worry about the possibility of differences in the versions of curl from system to system.  I can also retry endlessly and/or send out events from the script between retries.  And last, I intend to have different retry logic based upon whether the script is called interactively or not.

davehouse: sounds good. it is simpler to read like this too

PR #8, opened 2019-03-16
Bug 1510828 - convert remaining... 

markcor:
... Windows 10 hardware DSC configuration to Puppet configuration (except for generic-worker support).

Create profile for production critical Microsoft tools
- Add tools to win-packages module
Create profile and module for Microsoft services
- In this PR it effectively only adds KMS firewall exceptions, but this will grow for addtional KMS support and WSUS. 
Add support for UltraVNC 
- Including firewall restrictions for VNC and additionally SSH
- Add jumphost IPs to hiera
Create schedule task profile and module 
- Add maintain system task
-This will include additional tasks in the future 
Add support under network 
- Disable IPv6 explicitly 
- Set primary search domain 
Add support to completely disable all Windows Defender services
Expand support to disable Windows update
- Include support yo disable Win release that is in use in AWS
-- Add a custom fact to determine ID of specific OS release 
Disable Onedrive 
- Service and creation of schedule tasks
Add predefined Powershell profile
- Quality of life improvements for human interactions


modules/win_ultravnc/manifests/configuration.pp
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class win_ultravnc::configuration {
+
+    require win_ultravnc::install
+
+    file { $win_ultravnc::ini_file:
+        content => epp('win_ultravnc/ultravnc.ini.epp'),
```
dividehex: This file contains password hashes so make sure to add the show_diff => false attribute here

modules/win_ultravnc/templates/ultravnc.ini.epp
```diff
@@ -0,0 +1,71 @@
+[ultravnc]
+passwd=<%= $win_ultravnc::pw_hash %>
+<%# passwd2 is a default value. %>
+passwd2=DE864AF65F98840E2D
```
dividehex: AIUI, passwd2 is the 'read only' password, therefore it should also be looked up from hiera.

data/os/Windows.yaml
```diff
@@ -3,6 +3,13 @@
 # log aggregators
 win_dc_log_aggregator: "log-aggregator.srv.releng.%{facts.custom_win_mozspace}.mozilla.com"
 
+# jumphosts
+# Windows firewall exceptions uses IP addresses
+win_mdc1_jh1_ip:  "10.49.48.100" #"rejh1.srv.releng.mdc1.mozilla.com"
+win_mdc1_jh2_ip:  "10.49.48.101" #rejh2.srv.releng.mdc1.mozilla.com"
+win_mdc2_jh1_ip:  "10.51.48.100" #"rejh1.srv.releng.mdc2.mozilla.com"
+win_mdc2_jh2_ip:  "10.51.48.101" #"rejh2.srv.releng.mdc2.mozilla.com"
```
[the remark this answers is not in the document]
markcor: Puppets Windows Firewall module will only accept a string or undef for remote_ip. I tested an array and it failed.

markcor:
Addressed commits and reworked how to handle firewall on windows. This method will use define types that will have parameters passed to it through the profile. It will also set the default to block and will create reciprocal rules where needed.

markcor:
@dividehex 
- Removed the default block from the ssh and vnc profiles
- Correct tabbing in the firewall port defined types

PR #13, opened 2019-04-04
Mojave tweaks

dividehex:


modules/roles_profiles/manifests/profiles/cltbld_user.pp
```diff
@@ -14,12 +14,19 @@
...
                 password   => $password,
                 salt       => $salt,
                 iterations => $iterations,
             }
 
+            # Monkey patching directoryservice.rb in order to create users also breaks group merging
+            # So we directly add the user to the group(s)
+            exec { 'cltbld_developer_group':
+                command => '/usr/bin/dscl . -append /Groups/_developer GroupMembership cltbld',
+                unless  => '/usr/bin/groups cltbld | /usr/bin/grep -q -w _developer',
```
davehouse: you'd mentioned needing to do this. Is this maybe more maintainable than a patch anyway?

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker.pp
```diff
@@ -9,6 +9,10 @@
     case $::operatingsystem {
         'Darwin': {
 
+            class{ 'talos':
```
davehouse: nit. space

modules/macos_mobileconfig_profiles/manifests/desktop_background.pp
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class macos_mobileconfig_profiles::desktop_background {
+
+    mac_profiles_handler::manage { 'com.github.erikberglund.ProfileCreator.CB0ADD7C-78DE-4C29-854B-3FC332E953A6':
+        ensure      => 'present',
+        file_source => 'puppet:///modules/macos_mobileconfig_profiles/org.mozilla.desktop_background.mobileconfig',
```
davehouse: Is there another format for the profiles that could show the settings? I trust these, but I'm concerned we don't have a way to see what the profiles do when looking at them in the repo.

image

modules/roles_profiles/manifests/profiles/cltbld_user.pp
dividehex: The patch is still needed to create the users through puppet and I'm hoping puppetlabs will have the bug fixed soon.  We might be able to work some magic by signing the launchd provisioning script with our developer installation key if looks like the bug won't be fixed in a timely manner.  But for right now, this just works.

modules/macos_mobileconfig_profiles/manifests/desktop_background.pp
dividehex: I'm not sure if there is a cli tool to inspect them although they are binary files, you can open them up and see the payload.  The reason they are binaries is that I'm creating them the ProfileCreator tool and then signing them with our Developer Installer cert.

PR #15, opened 2019-04-05
Dump mac profile xml

davehouse:
This adds a prehook (local run of openssl) to dump out the MacOS profiles from binary to plist xml files. So we will have a direct view in the repo of what the profiles contain.

I'm guessing we could add a hook to re-sign the binary profiles when the xmls change also, but I don't think we'll be changing these often enough for that. And it might be simpler to put the plist/xml un-signed in another directory.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,9 @@ repos:
     name: macos-profile-xml
     description: Extract plist contents from signed MacOS profiles
     language: system
-    files: \.(.*)$
     verbose: True
-    pass_filenames: True
-    entry: bash -c 'for f in "$@"; do openssl smime -inform DER -verify -in "$f" -noverify -out "$f.xml" || echo "Failed extraction of $f"; done'
+    files: '\.mobileconfig$'
+    types: [binary]
+    entry: /bin/bash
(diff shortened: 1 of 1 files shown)


davehouse:
@dividehex I'd like to have the plain text version of these profiles in the repo. Would a prehook that creates those work for you? Is there a better location for the text versions, or should they be "*.plist" instead? Not all of the ones in the repo are signed right now. So that difference makes the *.xml filenames weird.

dividehex:
> 
> 
> @dividehex I'd like to have the plain text version of these profiles in the repo. Would a prehook that creates those work for you? Is there a better location for the text versions, or should they be "*.plist" instead? Not all of the ones in the repo are signed right now. So that difference makes the *.xml filenames weird.

I'd prefer to keep it simple and just place a copy of the unsigned file right next to it in the dir.  I'm only signing them for the sake of the day when Apple does something to enforce signing and/or we are forced to use MDM to propagate and manage profiles.  As for the ones that are unsigned, we can go back and sign them just for consistency sake.

davehouse:
That works for me. I'll close this out then and not add the hook for the .mobileprofile files.

PR #17, opened 2019-04-17
Windows Generic-worker support and automation. 

markcor:
This Pull Request includes: 
Bootstrap script. See  provisioners/windows/README.md for additional information.  

Schedule task creation and mangament of maintainsystem.ps1. This script includes directory clean up, determining if the node should run Puppet ,updating the local Puppet repo, applying Puppet, determining if the Puppet apply was successful or not, and creation of flag file that the Generic-worker wrapper script waits on before starting. 

Installation and configuration of Generic-worker 13.0.2 and supporting software for Moonshot nodes. Including a cleaner hardware wrapper script. 

Changes to the grant_symlnk_access.pp and support script becuase the previouc interation was failing but wasn't causing Puppet to fail. The same for the mozprofilerprobe.mof installation. 

Rework of win_packages::win_zip_pkg to us 7zip. This is because the forge unzip types fail when Puppet is ran under the System account context. 

Commenting out sym link creation because of failures when Puppet is ran under the System account context.  See bug in manifests comments.  This may require writing something custom. 

Commenting out OpenSSH installation because of failures when Puppet is ran under the System account context.  See bug in manifests comments.  This may need to rewrite the installation Powershell script. 

Other small improvements and corrections.


markcor:
Also note that all references to "markcor" will be replace with "mozilla-platform-ops" before merging.

provisioners/windows/geckotwin1064ht-bootstrap.ps1
```diff
@@ -0,0 +1,288 @@
...
+    $env:RUBYLIB = "$env:programfiles\Puppet Labs\Puppet\lib"
+    $env:USERNAME = "Administrator"
+    $env:USERPROFILE = "$env:systemdrive\Users\Administrator"
+
+    Write-Log -message  ('{0} :: Installing Puppetfile .' -f $($MyInvocation.MyCommand.Name)) -severity 'DEBUG'
+    R10k puppetfile install  --moduledir=r10k_modules
+   	# Needs to be removed from path or a wrong puppet file will be used
+    $env:path = ($env:path.Split(';') | Where-Object { $_ -ne "$env:programfiles\Puppet Labs\Puppet\puppet\bin" }) -join ';'
+    Get-ChildItem -Path $logdir\*.log -Recurse | Move-Item -Destination $logdir\old -ErrorAction SilentlyContinue
+    puppet apply manifests\nodes\nodes.pp --test --modulepath=modules`;r10k_modules --hiera_config=hiera.yaml --logdest $logdir\$datetime-bootstrap-puppet.log
```
dividehex: Instead of manifests\nodes\nodes.pp as an entry point, I would suggest just 'manifests'.  That allows site.pp to be included and ensures there is no default node.  Otherwise, there will be no toplevel scope. I would also remove --test and be explicit with what --test implies. eg --verbose, etc.

markcor:
Also had to update Tooltool source location because of https://github.com/mozilla/build-tooltool/pull/47.

PR #21, opened 2019-04-24
Bug 1545201 - [Ronin Majave] ssh module

dragoscrisan:
Bug 1545201 - [Ronin Majave] ssh module


modules/roles_profiles/manifests/roles/gecko_t_osx_1014.pp
```diff
@@ -22,4 +22,5 @@
     include ::roles_profiles::profiles::homebrew
     include ::roles_profiles::profiles::relops_users
     include ::roles_profiles::profiles::gecko_t_osx_1014_generic_worker
+    include ::roles_profiles::profiles::ssh
```
davehouse: Are there any risks in affecting the test pool by turning this on here? Maybe this PR could add the module, and a separate one could make the talos workers use it?

dividehex: I agree; add the module first and then include it in the role after testing.  As it is this module is untested and would definitely have broken puppet in production.

modules/ssh/files/allusers.bat
```diff
@@ -0,0 +1,12 @@
+REM This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: The scope of this the ssh module should be posix only since windows ssh is handled in another module.  In fact, the scope of this module for right now should be Mojave only.  Therefore remove this file and anything else that does windows or linux stuff.

modules/ssh/files/genkeys.rb
```diff
@@ -0,0 +1,41 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: remove this file.  It is not needed

modules/ssh/manifests/config.pp
```diff
@@ -0,0 +1,117 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh::config {
+    include users::root
+    include ssh::settings
+    include ssh::service
+    include ::config
+
+    case $::operatingsystem {
```
dividehex: windows should not be handled here.  Scope to mojave only.

modules/ssh/manifests/config.pp
```diff
@@ -0,0 +1,117 @@
...
+                        {
+                          identity => 'SYSTEM',
+                          rights   => ['full'] },
+                    ];
+            }
+        }
+
+        default: {
+            # Duo MFA requires special sshd configuration
+            $sshd_config_content = $duo_enabled ? {
```
dividehex: We aren't utilizing duo in this repo and when we do, we won't be using node scope vars.  So remove this.

modules/ssh/manifests/extra_authorized_key.pp
```diff
@@ -0,0 +1,19 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+define ssh::extra_authorized_key($username, $key, $home, $from='', $command='') {
```
dividehex: resource definitions and classes that utilize parameters should ALWAYS include there explicit type.  eg. String, Integer, Optional, etc.  Each parameter should also be on its own line.

modules/ssh/manifests/extra_authorized_key.pp
```diff
@@ -0,0 +1,19 @@
...
+    include ssh::settings
+
+    $authorized_keys = $::operatingsystem ? {
+        windows => "${ssh::settings::genkey_dir}/${username}.keys",
```
dividehex: Scope to mojave only!

modules/ssh/manifests/init.pp
```diff
@@ -0,0 +1,21 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh {
+    anchor {
```
dividehex: Anchors are deprecated in puppet 6.  You should utilize 'contains' instead.

modules/ssh/manifests/keys.pp
```diff
@@ -0,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh::keys {
+    include ::config
+
+    $from_extsync = hiera('ssh-keys', {})
```
dividehex: No Hiera lookups in component modules.  All lookups should be done within the profiles.  We also don't use extsync in ronin puppet.

modules/ssh/manifests/service.pp
```diff
@@ -0,0 +1,29 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh::service {
+    case $::operatingsystem {
+        Darwin : {
+            exec {
```
dividehex: -w is no longer valid in Mojave.  Since ssh is enabled in the deploystudio workflow, we don't need to use an exec here.  Simply use the service resource.

modules/ssh/manifests/service.pp
```diff
@@ -0,0 +1,29 @@
...
+                # command does tick the box for remote-login in the Sharing
+                # prefpane (survives reboot)
+                'turn-on-ssh' :
+                    command =>
+                    '/bin/launchctl load -w /System/Library/LaunchDaemons/ssh.plist',
+                    unless  =>
+                    "/usr/sbin/netstat -na | /usr/bin/grep -q 'tcp4.*\\*.22.*LISTEN'";
+            }
+
+            # Delete the com.apple.access_ssh group.  If present, this group limits
```
dividehex: We definitely shouldn't be deleting the ssh access group.  Instead, we should be adding them to the group.  I believe this already happens in the user module.

modules/ssh/manifests/user_privkey.pp
```diff
@@ -0,0 +1,17 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+define ssh::user_privkey($home, $username, $group, $key) {
```
dividehex: See previous comments about parameters.

modules/ssh/manifests/user_privkey.pp
```diff
@@ -0,0 +1,17 @@
...
+    $key_name = $title
+
+    file {
+        "${home}/.ssh/${key_name}":
+            ensure    => file,
+            mode      => filemode(0600),
```
dividehex: filemode is a puppetagain only lib.  Don't use it.

modules/ssh/manifests/userconfig.pp
```diff
@@ -0,0 +1,122 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+define ssh::userconfig($home='', $config='', $group='', $cleartext_password='',
```
dividehex: See previous comments about parameters.

modules/ssh/manifests/userconfig.pp
```diff
@@ -0,0 +1,122 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: This entire file is a whole lot of cruft.  It is the type stuff don't want to bring over from puppetagin.  And besides, user keys are already handled in the 'user' module.

https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/modules/users/manifests/user_ssh_config.pp

modules/ssh/templates/kts.ini.erb
```diff
@@ -0,0 +1,199 @@
+[KRegkey]
```
dividehex: Remove this file.

modules/ssh/templates/ssh_authorized_keys.erb
```diff
@@ -0,0 +1,13 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: ssh keys should be handled by the user modules.

modules/ssh/templates/extra_authorized_key.erb
```diff
@@ -0,0 +1,19 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: Ssh keys should be handled by the user module

modules/ssh/templates/sshd_config_duo.erb
```diff
@@ -0,0 +1,29 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: Remove this file.  It is out of scope at this time.

modules/ssh/templates/sshd_config.erb
```diff
@@ -0,0 +1,28 @@
...
+AcceptEnv XMODIFIERS
+
+LogLevel VERBOSE
+
+<% if @only_user_ssh -%>
+PermitRootLogin no
+PasswordAuthentication no
+PermitEmptyPasswords no
+<% else -%>
+PasswordAuthentication yes
```
dividehex: Although password authentication is a bad thing, we will need it for the time being.

modules/ssh/manifests/settings.pp
```diff
@@ -0,0 +1,17 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh::settings {
+    # config files' paths are different per platform
+    case $::operatingsystem {
+        Darwin: {
+            $ssh_config           = '/etc/ssh_config'
+            $sshd_config          = '/etc/sshd_config'
+            $known_hosts          = '/etc/ssh_known_hosts'
+            $notify_on_key_change = []
```
dividehex: What is this used for?  I bet it is unnecessary.

modules/ssh/manifests/config.pp
```diff
@@ -0,0 +1,117 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class ssh::config {
+    include users::root
```
dividehex: There is no users::root.

modules/ssh/manifests/config.pp
```diff
@@ -0,0 +1,117 @@
...
+    include ssh::settings
+    include ssh::service
+    include ::config
```
dividehex: There is no ::config module.

modules/ssh/templates/sshd_config.erb
```diff
@@ -0,0 +1,28 @@
...
+
+# Accept locale-related environment variables
+AcceptEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
+AcceptEnv LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT
+AcceptEnv LC_IDENTIFICATION LC_ALL LANGUAGE
+AcceptEnv XMODIFIERS
+
+LogLevel VERBOSE
+
+<% if @only_user_ssh -%>
```
dividehex: only_user_ssh is a node scope var in puppet again.  There will be no node scope vars in ronin puppet.  There will also be no root user account logins PERIOD.  Therefore, it should always be 'PermitRootLogin no'

davehouse:
there may be a good simple module in the puppet forge for ssh keys. I think I saw the puppetlabs account module can manage ssh keys

davehouse:
> there may be a good simple module in the puppet forge for ssh keys. I think I saw the puppetlabs account module can manage ssh keys

dividehex pointed out that the existing ronin users module already manages ssh keys. So that part is not needed.

PR #22, opened 2019-04-25
add bitbar_devicepool module

aerickson:
Also adds test-kitchen integration testing. See README.md for more information.

`provisioners/linux/bootstrap_bitbar_devicepool.sh` is mostly taken from `provisioners/mac/bootstrap_mojave.sh` and `modules/puppet/templates/puppet-darwin-run-puppet.sh.erb`. It can be used as a run script also.


provisioners/linux/bootstrap_bitbar_devicepool.sh
```diff
@@ -0,0 +1,151 @@
...
+R10K_DIR="/etc/puppetlabs/environments/production/r10k_modules"
+ROLE_FILE='/etc/puppet_role'
+PUPPET_REPO_PATH="$PUPPET_ENV_DIR/production"
+
+FQDN=$(${FACTER_BIN} networking.fqdn)
+
+
+## FUNCTIONS
+
+function update_puppet {
```
davehouse: would it be worth sharing functions across the platforms, or merging the bootstrap scripts? This looks similar to the mojave bootstrap, and I'm guessing the linux-talos one will match closely.

test/integration/bitbar/serverspec/bitbar_devicepool_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe command('cd /home/bitbar/mozilla-bitbar-devicepool && git status') do
+    its(:exit_status) { should eq 0 }
+  end
+end
+
+describe 'service' do
+  # service:is_installed is windows only
+  describe command('systemctl status bitbar') do
+    # code 3 is loaded, but not running
+    its(:exit_status) { should eq 3 }
```
davehouse: This is great; I like having tests! I'm always needing to check state manually when making changes in puppet, and putting those tests into reproducible tests is excellent.

README.md
```diff
@@ -1,3 +1,49 @@
...
+```
+modules/roles_profiles/manifests/roles
+modules/roles_profiles/manifests/profiles
+```
+
+Roles specify everything a machine type needs to fufill a role.
+
+Profiles provide an OS-independent interface to functionality provided by modules.
+
+## testing
```
davehouse: thankyou for the documentation. I know nothing about kitchen, and this shows me how to start using it

manifests-kitchen/bitbar.pp
```diff
@@ -0,0 +1,29 @@
+##################################################
+#### MOCK CLASSES WHICH SHOULD NOT TESTED HERE
+# class kitchen_template2(
+#   Hash $config = {},
+# ) {
+#   notice( 'mocked class ==> kitchen_template::foobar' )
+# }
+
+# dirty hack: include site.pp
+# can't point at directory of manifests, as last check in site.pp fails for test-kitchen
```
davehouse: is this an issue with kitchen, or with our puppet? (can we create an issue to fix the cause or is it unlikely to affect other platforms?)

manifests/site.pp
```diff
@@ -2,6 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# NOTE(aerickson): please don't add more global variables.
+#   variables like this should be set in profile modules.
+#   these make testing difficult (this block must be
+#   copy/pasted into every puppet-kitchen manifests).
```
davehouse: What if we remove these and put it into a profile?

manifests-kitchen/bitbar.pp
aerickson: I don't think puppet should have deprecated `import` (the recommend alternative is to use include (but that only works for modules) or to use a directory of manifests). In normal puppet use cases, you don't need it, but wIth puppet-kitchen it would be very useful to compose manifests.

I've figured out a workaround using symlinks. :)
I can't think of a fix we could ask for upstream (other than re-enabling import). We could refactor our code to not use these globals... but I guess it's the only way to set 

I think we should settings like this should live in profiles (or hiera if we really want to set something globally).

manifests/site.pp
aerickson: We could, but then we'd have to set them in each profile... I see the power that the globals afford us here.

test/integration/bitbar/serverspec/bitbar_devicepool_spec.rb
aerickson: Yeah, the cycle of converge and manual check things out gets old. :)

README.md
aerickson: For sure. :D

provisioners/linux/bootstrap_bitbar_devicepool.sh
aerickson: The mojave bootstrap script is oriented at bootstrapping and then cleaning up (it downloads a zip of the repo and removes on exit). There's also a run script generated for OS X. 

The bitbar script is a bit of a hybrid. It clones a repo and is designed to be a run script also. 

There are definitely parts we could pull out and reuse between them.

Part of what's nice about the scripts is that they're self contained (it's a single file to scp over) and we'd lose that if we broke stuff out (unless we wrote a simple script generator the prepended the common bits).

davehouse: That sounds reasonable. If we start duplicating to 5 or so of these scripts that are so similar, then I'd like to combine them, but even then the conditionals might be ugly.

davehouse:
+1 I tested the vagrant run and (changing the bootstrap to use this branch) it worked as described in the readme.

PR #23, opened 2019-04-29
Bug 1546448 notarization base

davehouse:
Set up role for macos mojave notarization workers

1. base role and profile
2. remove bootstrap user (only users with keys can login)
3. add users+keys for signing admins


aerickson:
lgtm r+

I think you'll need to rebase now that I've landed my bitbar-devicepool change.

davehouse:
@aerickson thx. I updated my branch. Could I get your re-review?

PR #26, opened 2019-05-02
Bug 1546827 - get and add cert for https://tooltool.mozilla-releng.net

markcor:
Without it tooltool will fail on the SSL connection. 

Additionally this pull request:
Adds support to manage the relengapi.tok.
Moves cacert.pem tot he proper directory.
Adds sublimetext to the catalog 
Corrects the location of where to move old generic-worker logs. 
Adds a line to maintainsystem.ps1 to delete logs older than 7 days.


modules/win_mozilla_build/manifests/tooltool.pp
```diff
@@ -29,4 +32,16 @@
                                         },
         inherit_parent_permissions => true,
     }
+    file { "${builds}\\relengapi.tok":
+        content   => $win_mozilla_build::tooltool_tok,
+        show_diff => false,
+    }
+    # This script will download and add the certifictae form https://tooltool.mozilla-releng.net
+    # Without tooltool will fail during tasks
```
dividehex: I'm not sure what the comments mean here.

markcor: That was very poorly worded. It is now updated.

PR #29, opened 2019-05-03
 Bug1546821 - Initial win ux worker support 

markcor:
Add roles for UX laptops.
Format provisioners/windows/README.md


davehouse:
+1

PR #32, opened 2019-05-10
[WIP] telegraf setup for windows hardware

davehouse:


davehouse:
instead of adding telegraf on windows, we may be able to add metrics collection into nxlog

PR #33, opened 2019-05-13
Bug1550848 worker python deps

davehouse:
mercurial plugin robust-checkout was not being placed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1550848


davehouse:
Thanks!

PR #37, opened 2019-05-17
Bug1523744 macos logging

davehouse:
* collect logs from macos unified logging
* forward logs to stackdriver

1. installs fluentd (td-agent) from dmg in s3
2. installs stackdriver plugin (and required coreutils from homebrew)
3. configures to collect logs from apple's `log stream` commandline tool
4. configures forwarding logs to stackdriver
  4a. adds auth for stackdriver service account to write logs


modules/fluentd/templates/fluentd.conf.erb
```diff
@@ -0,0 +1,84 @@
+<source>
+  @type exec
+  command "log stream --color none --level debug --type log --style syslog \
+      --timeout 1d \
+    | tail -n +3 \
+    | awk '{printf (/^[0-9]{4}/&&NR>1?RS:x)\"%s\",$0} END {print \"\"}'"
+    # tail to omit first 2 lines that cause syslog parsing to fail:
+    #   Filtering the log data using "type == 1024"
+    #   Timestamp                       (process)[PID]    
+    # awk remove newlines apple's log output as syslog prints in log messages
```
davehouse: To simplify this (removing the step with awk), we could output json and capture specific fields. However, Json input for fluentd expects to not have the entries in an array: some string manipulation will still be required.

modules/fluentd/manifests/init.pp
```diff
@@ -0,0 +1,61 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class fluentd (
+    String $worker_type,
+    String $stackdriver_project,
+) {
+
+    $stackdriver_keyid    = lookup("stackdriver.${stackdriver_project}.keyid")
```
davehouse: I added these secrets to the vault.yaml on install2 mdc1/2

modules/fluentd/templates/fluentd.conf.erb
davehouse: I switched to the default log style to get the log level/severity.

modules/fluentd/manifests/init.pp
dividehex: As we discussed over video, I'd like to see these lookups done from a profile and passed as parameters to the fluentd class.  A generic 'logging' profile should be used for all roles.

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker.pp
```diff
@@ -6,11 +6,18 @@
...
     require roles_profiles::profiles::cltbld_user
 
+    $worker_type = 'gecko-t-osx-1014'
+
     case $::operatingsystem {
         'Darwin': {
 
             include puppet::atboot
 
+            class { 'fluentd':
```
dividehex: Move this to a generic 'logging' profile and then include it on the gecko 1014 role.

davehouse:
t-mojave-r7-467.test.releng.mdc1.mozilla.com is running with this config

davehouse:
Thank you! I'll reference your note for a revision/update

PR #41, opened 2019-05-22
Mac logging fix

davehouse:


modules/packages/manifests/fluent_plugin_google_cloud.pp
```diff
@@ -8,10 +8,7 @@
     $version = '0.7.11'
 
     # fluent-plugin-google-cloud requires json which expects coreutils (requires gnu mkdir)
-    package { 'coreutils':
-        ensure   => present,
-        provider => brew,
-    }
+    require packages::coreutils
```
davehouse: with only the package entry for coreutils, the following exec for installing the plugin was executed before coreutils was installed.

by moving it to a module required in the homebrew profile (I am assuming that like xcode_cmd_line_tools, coreutils is a base package other tools will need), it gets installed before the fluent-plugin-google gem

davehouse: adding a require here, I think makes sure that it is available (and shows that it is required, for when we read this in a year)

PR #42, opened 2019-05-23
Puppet apply email reports and telegraf metrics

dividehex:
The puppet atboot now fires off a metric tick when puppet is successful or fails.  On failure, we also fire off an email report to a google groups email.

### DIFF
--- a/modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
+++ b/modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
@@ -145,6 +145,7 @@ function run_puppet {
         retval=1
     fi
 
+    LOG_OUT=$(cat "{$TMP_LOG}")
     rm "${TMP_LOG}"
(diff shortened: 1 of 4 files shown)


modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
```diff
@@ -68,6 +68,38 @@ function block_on_network {
...
+   smtpObj.sendmail("${SENDER}", "${RECEIVER}", msg)
+   print "Successfully sent email"
+except SMTPException:
+   print "Error: unable to send email"
+EOF
+}
+
+function notify_telegraf {
+    TELEGRAF_TABLE=$1
+    META_DATA="<% @meta_data.each do |key,value| -%>,<%= key -%>=<%= value -%><% end -%>"
```
davehouse: nice way of putting in the tags with the leading commas!

PR #43, opened 2019-05-23
[WIP] Notarization mdc2

davehouse:
troubleshooting set up for notarization servers in mdc2


davehouse:
remote-hands will re-image the machines
with the puppet-mail and metrics, I can check if puppet failed (problems logging in after the first reimage)

https://groups.google.com/a/mozilla.com/forum/#!forum/puppet-ronin-reports

PR #49, opened 2019-05-29
Draft: Signing worker python environment

srfraser:
This still requires work on the configuration file template, but it would be good to get a first pass review / CI check on the contents.

I've had to import the virtualenv.pp file from puppet/python module as it required modification to work on Macs - the original is Linux-only.


modules/packages/manifests/p7zip.pp
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class packages::p7zip {
```
escapewindow: We no longer need 7zip, so we can drop this :)

modules/packages/manifests/supervisor.pp
```diff
@@ -0,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class packages::supervisor {
```
escapewindow: If we can get supervisor+python from s3, we may no longer have a homebrew dep, and we won't be blocked from deploying to beta/esr68.

modules/signing_worker/files/requirements.txt
```diff
@@ -0,0 +1,2 @@
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=scriptworker_client&subdirectory=scriptworker_client
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=iscript&subdirectory=iscript
```
escapewindow: I added the vendored/mozbuild dir as a requirement.
I'm guessing we may want to copy all of these non-pypi modules up to s3 once they're in a released state.

modules/signing_worker/templates/scriptworker_config.erb
```diff
@@ -0,0 +1,30 @@
...
+        keychain_password: ...
+        pkg_cert_id: ...
+        apple_notarization_account: ...
+        apple_notarization_password: ...
+        concurrency_limit: 10
+        notarization_poll_timeout: 900
+        widevine_url: ...
+        widevine_user: ...
+        widevine_pass: ...
+        widevine_cert: ...
```
escapewindow: i added `omnija_{url,user,pass}`.

modules/signing_worker/templates/scriptworker_config.erb
```diff
@@ -0,0 +1,30 @@
+work_dir: /tmp/work
+artifact_dir: /tmp/artifacts
+taskcluster_scope_prefix: "project:releng:signing:"
+verbose: true
+local_notarization_accounts: ["account1"]
+concurrency_limit: 2
+7z: /usr/ocal/bin/7z
+mac_config:
+    dep:
```
escapewindow: We'll also need to populate `nightly` and `release` blocks in `mac_config`, which look like `dep` but different.

modules/signing_worker/templates/scriptworker_config.erb
```diff
@@ -0,0 +1,30 @@
...
+        notarize_type: multi_account
+        signing_keychain: /path/to/dep.keychain
+        supported_behaviors:
+            - mac_pkg
+            - mac_sign
+            - mac_sign_and_pkg
+            - mac_notarize
```
escapewindow: `dep` will not support `mac_notarize` when we roll out.

modules/signing_worker/templates/scriptworker_config.erb
```diff
@@ -0,0 +1,30 @@
...
+        sign_with_entitlements: true
```
escapewindow: I *think* `dep` will not sign with entitlements when we roll out.

modules/packages/manifests/p7zip.pp
srfraser: Yeah, I'd removed the installation but had left the packages manifest in case it was useful elsewhere.

modules/packages/manifests/supervisor.pp
srfraser: The awkwardness here is setting up supervisord to run. As mentioned in the channel, this either means puppet runs regularly (which I don't think it will) to ensure it's still running, or setting up launchd ourselves, which homebrew is currently doing for us. It's a step backwards to have to roll all this ourselves. 
It also reopens the question of whether we just use launchd directly for the scriptworker.

modules/signing_worker/templates/scriptworker_config.erb
srfraser: I'm not too concerned at the moment since the entire config file will need rewriting once the secrets service is available

modules/roles_profiles/manifests/profiles/notarization_users.pp
```diff
@@ -0,0 +1,37 @@
...
+
+    case $::operatingsystem {
+        'Darwin': {
+            # Make sure the users profile is required
+            # That is where the user virtual resources are generated
+            require roles_profiles::profiles::users
+            # Lookup the group array and realize their user resource
+            $users = lookup('user_groups.notarization', Array, undef, undef)
+            realize(Users::Single_user[$users])
+
```
escapewindow: `/usr/sbin/sysadminctl -addUser notary10 -admin -password -` as root worked for me manually on mac-v3-signing2.mdc1. We could possibly add the password to the commandline if needed.

escapewindow:
We need to add the self-signed developer ID CA, a la https://github.com/mozilla-releng/build-puppet/blob/master/modules/signingserver/manifests/base.pp#L51-L65

escapewindow:
We also probably want to add the log-aggregator / mig stuff from build-puppet, so we get login alerts.

escapewindow:
I tested dep -- we can have multiple daemons if we have multiple users.

So for dep,

- no notary users
- depbld{1,2} ?
    - these need sudo and a scriptworker venv per
    - the script_config.yaml's and scriptworker.yaml's will be slightly different as well.

We may end up manually configuring the pool during the first pass to be able to roll out this week. We can potentially quarantine a single host and test our puppet configs, then roll out when we're confident.

escapewindow:
https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=34014655 for how we reimaged the boxes.

https://github.com/escapewindow/scriptworker-scripts/wiki/manual-rollout for how we rolled out the pools manually.

srfraser:
Closing to rework and clear dodgy history

PR #54, opened 2019-06-11
Add windows md files

markcor:
These are not the final version of the files. However, I wanted to get something up since these are now in production.


markcor:
Also note that there will be some more md files for very specific items.

PR #67, opened 2019-07-10
Update macos desktop backgroud profile to allow user changes

dividehex:


dividehex:
https://bugzilla.mozilla.org/show_bug.cgi?id=1549046

PR #70, opened 2019-07-30
Bug 1559443 - Additional Support for AWS Windows 

markcor:
This pull request includes: microsoft_network_services

Bug 1562040 -  roles_profiles::profiles::
Adding MDC 1 KMSserver IP and KMS key (not a private value) to Windows.yaml.
Generation of custom facts to determine OS caption and verify if activation is needed. 
Creation of win_kms::force_activation to create a Powershell script and pass values specified in the profile. 

Bug 1562969 - roles_profiles::profiles::ec2_instance_configuration
Management of the EC2 config file 

Bug 1563289 - roles_profiles::profiles::ec2_instance_configuration
Setting Instance name and search domain. Required creation of custom facts to base values off of instance metadata. 

Bug 1562974 - roles_profiles::profiles::files_system_managment 
Management of page file. 

Bug 1562038 - roles_profiles::profiles::firewall
Fix disabling firewall for AWS instances and only do it if the firewall is enabled. 

Bug 1563287 - roles_profiles::profiles::network
Set network category to private for AWS instances. 

Bug 1544141 - roles_profiles::profiles::ssh 
Fix OpenSSH install (not install on production workers at this time)

Removal of modules/win_shared/facts.d/win_custom_os_facts.ps1 which was a duplication of  facts_win_custom_os.ps1.


modules/roles_profiles/manifests/profiles/network.pp
```diff
@@ -10,8 +10,22 @@
...
         }
         'Windows': {
-            include win_network::set_search_domain
-            include win_network::disable_ipv6
+
+            if $facts['custom_win_location'] == 'datacenter' {
+                include win_network::set_search_domain
+                include win_network::disable_ipv6
+            }
+            if $facts['custom_win_location'] != 'datacenter' {
```
dividehex: This should be an `else` so you aren't evaluating twice

markcor: Makes sense. I will change it in a future patch, so not to clear out this review.

PR #71, opened 2019-07-31
Stop macos mojave welcome

davehouse:
prevent the "welcome to macos mojave" notification


davehouse:
this did not work.
I did not find a way to disable the welcome notification (these only disabled the view of the tour .. so the notification link did not load the tour :/)

PR #72, opened 2019-07-31
Add gecko-3-t-osx-1014 role/profile

dividehex:


tomprince:
These workers will need stable ed255119 CoT (which is high-value), the public key of which will need to get added to https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/constants.py#L94 and at least https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/modules/generic_worker/templates/generic-worker.config.erb#L27 will need to be updated for these workers as well.

PR #73, opened 2019-08-02
Mac signing scriptworkers

srfraser:
This will need secrets updating in the vault, how should we go about that?


escapewindow:
Nice! Looks like the failures are formatting related.

```
modules/signing_worker/manifests/base.pp - WARNING: indentation of => is not properly aligned (expected in column 19, but found it in column 17) on line 126
874
modules/signing_worker/manifests/base.pp - WARNING: indentation of => is not properly aligned (expected in column 19, but found it in column 21) on line 127
```

modules/roles_profiles/manifests/profiles/notarization_users.pp
```diff
@@ -0,0 +1,54 @@
...
+            include sudo
+            # Make sure the users profile is required
+            # That is where the user virtual resources are generated
+            require roles_profiles::profiles::users
+            # Lookup the group array and realize their user resource
+            $users = $::hostname ? {
+                /^(tb-)?mac-v3-signing\d+/ =>
+                    lookup('user_groups.notarization', Array, undef, []),
+                /^dep-mac-v3-signing\d+/ =>
+                    lookup('user_groups.dep_notarization', Array, undef, undef),
```
escapewindow: Is this a preferred method over having a settings.pp with the various settings specified per group/trustlevel, then specifying the group or trustlevel via hostname regex?

modules/signing_worker/files/requirements.txt
```diff
@@ -0,0 +1,5 @@
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=mozbuild&subdirectory=vendored/mozbuild
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=scriptworker_client&subdirectory=scriptworker_client
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=iscript&subdirectory=iscript
+scriptworker
+# TODO add widevine once we have a wheel or private repo solution
```
escapewindow: Until we have some solution for widevine (github secret for cloning a private repo?) we'll still have at least one manual step before we can go from imaged/puppetized to live.

modules/signing_worker/manifests/base.pp
```diff
@@ -0,0 +1,129 @@
...
+    $certs_dir = "${signing_worker::scriptworker_base}/certs"
+    $tmp_requirements = "${signing_worker::scriptworker_base}/requirements.txt"
+    $scriptworker_config_file = "${signing_worker::scriptworker_base}/scriptworker.yaml"
+    $script_config_file = "${signing_worker::scriptworker_base}/script_config.yaml"
+    $scriptworker_wrapper = "${signing_worker::scriptworker_base}/scriptworker_wrapper.sh"
+
+    # Dep workers have a non-deterministic suffix
+    $worker_id = "${::hostname}${signing_worker::worker_id_suffix}"
+
+    $role = $::hostname? {
```
escapewindow: More hostname configs. I'm thinking a settings.pp may be cleaner.

modules/signing_worker/templates/dep_script_config.yaml.erb
```diff
@@ -0,0 +1,34 @@
...
+        apple_notarization_password: "<%= @notarization_password %>"
+        apple_asc_provider: <%= @apple_asc_provider %>
+        notarization_poll_timeout: <%= @notarization_poll_timeout %>
+        widevine_url: <%= @dep_widevine_url %>
+        widevine_user: <%= @dep_widevine_user %>
+        widevine_pass: <%= @dep_widevine_pass %>
+        widevine_cert: <%= @dep_widevine_cert %>
+        omnija_url: <%= @dep_omnija_url %>
+        omnija_user: <%= @dep_omnija_user %>
+        omnija_pass: <%= @dep_omnija_pass %>
```
escapewindow: @Callek 's adding langpack support to iscript. Let's have you two talk before we roll out so we don't blow away his manual changes.

modules/signing_worker/manifests/base.pp
```diff
@@ -0,0 +1,129 @@
...
+        owner   => $signing_worker::user,
+        group   => $signing_worker::group,
+    }
+
+    $launchd_script = "/Library/LaunchDaemons/org.mozilla.scriptworker.${signing_worker::user}.plist"
+    file { $launchd_script:
+        content => template('signing_worker/org.mozilla.scriptworker.plist.erb'),
+        mode    => '0644',
+    }
+    exec { "${signing_worker::user}_launchctl_load":
```
escapewindow: We may want to be able to perform some checks before starting up. Should we have two levels of scripts (a check and a startup wrapper), or should we put those checks into the wrapper?

srfraser:
> Overall, awesome! Thank you!
> 
> We need:
> 
>     * a secrets solution :)

Out of scope for this PR, I think
 
>     * a widevine solution, or block startup on a human cloning widevine and installing

I'm not sure how puppet could clone a private repo

>     * @Callek 's langpack changes that he'll be rolling out shortly

I don't know about these, what are the changes?

> 
>     * I think we'll regret not having a settings.pp with the various configs that are currently governed by hostname regexes all over

If it was going to expand, I'd agree, but we'd just be removing the regular expressions out. Given the special nature of the dep hosts I'm not sure it's easily configurable that way.

modules/signing_worker/manifests/base.pp
srfraser: What sort of checks? The startup wrapper will keep being run over and over until it stays running, so it depends how the checks should behave.

modules/roles_profiles/manifests/profiles/notarization_users.pp
srfraser: It's the method used for our own accounts, which are arguably the same importance.

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -32,15 +32,42 @@
...
+            class { 'scriptworker_prereqs': }
 
-            contain packages::python3_s3
-            file { '/tools/python3':
-                    ensure  => 'link',
-                    target  => '/usr/local/bin/python3',
-                    require => Class['packages::python3_s3'],
-            }
-            contain packages::virtualenv_python3_s3
+            case $::hostname {
```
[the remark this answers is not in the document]
srfraser: There's no central nodes.pp file, one gets generated by the bootstrapping, so we shouldn't have one committed. So that means moving things into the roles.

modules/python/manifests/virtualenv.pp
```diff
@@ -0,0 +1,142 @@
...
+        mode   => $mode,
+      }
+    }
+
+    $pip_cmd   = "${venv_dir}/bin/pip"
+    $pip_flags = "${pypi_index} ${proxy_flag} ${pip_args}"
+
+    # Unless activate exists and VIRTUAL_ENV is correct we re-create the virtualenv
+    exec { "python_virtualenv_${venv_dir}":
+      command     => @("COMMAND"/L),
```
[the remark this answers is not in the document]
srfraser: It's a puppet equivalent of `cat <<EOF`, essentially. The command to run was too long for linter happiness, so to split it up we do this, and it reads until `|-COMMAND`, below.

modules/python/manifests/virtualenv.pp
```diff
@@ -0,0 +1,142 @@
...
+        true ${proxy_command} && \
+        ${virtualenv_cmd} ${system_pkgs_flag} -p ${python} ${venv_dir} && \
+        ${pip_cmd} --log ${venv_dir}/pip.log install ${pip_flags} --upgrade pip
+        |-COMMAND
+      user        => $owner,
+      creates     => "${venv_dir}/bin/activate",
+      path        => $path,
+      cwd         => '/tmp',
+      environment => $environment,
+      unless      => "grep '^[\\t ]*VIRTUAL_ENV=[\\\\'\\\"]*${venv_dir}[\\\"\\\\'][\\t ]*$' ${venv_dir}/bin/activate",
```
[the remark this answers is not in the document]
srfraser: Thankfully plagarised from the module attributed at the top!

modules/signing_worker/files/requirements.txt
srfraser: If there's a manual step I should remove the `launchctl load` directive and have someone do that manually.

escapewindow:
> > Overall, awesome! Thank you!
> > We need:
> > ```
> > * a secrets solution :)
> > ```
> 
> Out of scope for this PR, I think

We need it before we can use this PR to image our machines, no? And since Relops uses ronin-puppet to half-puppetize our machines after imaging, if we land this and don't have the secrets to populate, we may cause problems.

Because of the above, I believe we need to have the secrets solution before we can land this. I may be wrong.

> > ```
> > * a widevine solution, or block startup on a human cloning widevine and installing
> > ```
> 
> I'm not sure how puppet could clone a private repo

a) land ssh key somewhere, b) use ssh key to `git clone` ?

> > ```
> > * @Callek 's langpack changes that he'll be rolling out shortly
> > ```
> 
> I don't know about these, what are the changes?

Probably around https://github.com/mozilla-releng/scriptworker-scripts/pull/11#pullrequestreview-270950679 .

> > ```
> > * I think we'll regret not having a settings.pp with the various configs that are currently governed by hostname regexes all over
> > ```
> 
> If it was going to expand, I'd agree, but we'd just be removing the regular expressions out. Given the special nature of the dep hosts I'm not sure it's easily configurable that way.

We will be adjusting these settings in the future. As it stands, we'll need to find all files that are applicable and adjust them, as opposed to adjusting the settings.pp file.

modules/signing_worker/manifests/base.pp
escapewindow: e.g., don't start up if the widevine module isn't installed.

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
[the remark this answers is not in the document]
escapewindow: Ok. I still think a settings.pp would be preferable, though we may have to keep the hostname regexes in the roles.

srfraser:
> > > * a secrets solution :)

> > Out of scope for this PR, I think
> 
> We need it before we can use this PR to image our machines, no? And since Relops uses ronin-puppet to half-puppetize our machines after imaging, if we land this and don't have the secrets to populate, we may cause problems.

There's currently a set of hiera secrets that are copied out of vault and onto the node itself for deployment. Is that enough, or does it need more? Because I don't think I can do more without relops deploying something.

> > I'm not sure how puppet could clone a private repo
> 
> a) land ssh key somewhere, b) use ssh key to `git clone` ?

Urgh. Yes, but also urgh. This is getting very hacky.

> We will be adjusting these settings in the future. As it stands, we'll need to find all files that are applicable and adjust them, as opposed to adjusting the settings.pp file.

I'm not sure how the different users to be deployed would work, in that scenario, and the certificates (given we can't map hostname to value directly due to yaml restrictions)

modules/roles_profiles/manifests/profiles/notarization_users.pp
escapewindow: > > > > * a secrets solution :)
> 
> > > Out of scope for this PR, I think
> > 
> > 
> > We need it before we can use this PR to image our machines, no? And since Relops uses ronin-puppet to half-puppetize our machines after imaging, if we land this and don't have the secrets to populate, we may cause problems.
> 
> There's currently a set of hiera secrets that are copied out of vault and onto the node itself for deployment. Is that enough, or does it need more? Because I don't think I can do more without relops deploying something.

If we have a way of updating that set of hiera secrets that gets deployed onto the node, I consider that a secrets solution.

> > > I'm not sure how puppet could clone a private repo
> > 
> > 
> > a) land ssh key somewhere, b) use ssh key to `git clone` ?
> 
> Urgh. Yes, but also urgh. This is getting very hacky.

I think my preferred long term solution is a non-public s3-or-similar url where we publish the widevine wheel. As long as it's internal-only, we can download it and install it.

> > We will be adjusting these settings in the future. As it stands, we'll need to find all files that are applicable and adjust them, as opposed to adjusting the settings.pp file.
> 
> I'm not sure how the different users to be deployed would work, in that scenario, and the certificates (given we can't map hostname to value directly due to yaml restrictions)

[Here](https://github.com/mozilla-releng/build-puppet/blob/master/modules/signing_scriptworker/manifests/settings.pp#L31-L45) we have a nickname, `dep`, that specifies configuration. The users to be deployed could be a list in that dict, no? Then the hostname regex would only need to specify that `dep-mac...` is `dep` configuration, and we would have a single location to update our configs.

escapewindow:
Nick and I chatted. It's possible we won't be able to tweak the widevine access privs til Chris comes back. He also pointed out you will be out next week.

What do you think is a good state to merge? I'm guessing that we might leave some things for manual deployment, like widevine, possibly the keychains or config files, and update the [wiki](https://github.com/mozilla-releng/scriptworker-scripts/wiki/manual-rollout) with those steps. In that case, we might want to leave the launch daemons disabled, but hopefully allow for manually enabling them once a human has done the needed steps. I'm thinking it would be good to have these changes landed; my main concerns were a) are we breaking the puppet install if we're missing secrets in the relops store, or b) are we starting up scriptworker before we're ready to take tasks. We might also want some relops eyes on this, since I'm not a ronin-puppet expert.

modules/roles_profiles/manifests/profiles/notarization_users.pp
srfraser: That nickname is mostly for the templates at the moment, I think it would be confusing to overload that and then not use a lot of the variables in the templates. 

`dep` doesn't just get a list of users, but each user has its own suffix and directory. I can see what a config area looks like, but I suspect it will be harder to read.

srfraser:
@dividehex We have some secrets to get added to vault, how should we get those to you?

dividehex:
> @dividehex We have some secrets to get added to vault, how should we get those to you?
PGP or keybase.io

srfraser:
I've emailed the vault additions. This should be good to go for ff, tb and dep deployments, as long as someone logs in and deploys widevine manually.  Once we can get a read-only ssh key for that repository we can clone it during the puppet deploy, and uncomment the `launchctl load` directive.

Callek:
@srfraser status of this PR?

srfraser:
> @srfraser status of this PR?

Waiting on the secrets to be updated in the central store. After it's merged in we'll still need to deploy the certificates/keychains manually and then `launchctl load` the service, as those certificates needed more protection than config management could offer.

modules/signing_worker/templates/requirements.txt.erb
```diff
@@ -0,0 +1,5 @@
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=mozbuild&subdirectory=vendored/mozbuild
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=scriptworker_client&subdirectory=scriptworker_client
+-e git+https://github.com/escapewindow/scriptworker-scripts#egg=iscript&subdirectory=iscript
```
nthomas-mozilla: Could we pls update these to the mozilla-releng repo, thx,

modules/signing_worker/templates/script_config.yaml.erb
```diff
@@ -0,0 +1,50 @@
...
+dmg_prefix: "<%= scope['signing_worker::dmg_prefix'] %>"
+mac_config:
+<% @role_config.each do |role, role_data| -%>
+    <%= role %>:
+        notarize_type: multi_account
+        signing_keychain: "<%= @certs_dir %>/<%= role %>-signing.keychain"
+        supported_behaviors:
+            - mac_sign
+            - mac_sign_and_pkg
+            - mac_pkg
```
escapewindow: We should `s,mac_pkg,mac_geckodriver,` on the prod mac signing scriptworkers.

modules/signing_worker/templates/dep_script_config.yaml.erb
```diff
@@ -0,0 +1,34 @@
...
+    - "/Library/Keychains/System.keychain"
+dmg_prefix: <%= @dmg_prefix %>
+mac_config:
+    dep:
+        notarize_type: multi_account
+        signing_keychain: <%= @base_dir %>/dep-signing.keychain
+        supported_behaviors:
+            - mac_sign
+            - mac_sign_and_pkg
+            - mac_pkg
```
escapewindow: I believe we can drop `mac_pkg`.

modules/signing_worker/manifests/base.pp
```diff
@@ -0,0 +1,173 @@
...
+    $notarization_users = $::hostname ? {
+        /^(tb-)?mac-v3-signing\d+/ =>
+            lookup('user_groups.notarization', Array, undef, []),
+        /^dep-mac-v3-signing\d+/ =>
+            lookup('user_groups.dep_notarization', Array, undef, []),
+        default => [],
+    }
+
+    $supported_behaviors = $::hostname ? {
+        /^(tb-)?mac-v3-signing\d+/ => ['mac_sign', 'mac_sign_and_pkg', 'mac_geckoview', 'mac_notarize'],
```
escapewindow: I think this should be `s,mac_geckoview,mac_geckodriver,`

srfraser:
> Awesome! And thank you for the [documentation](https://github.com/mozilla-releng/scriptworker-scripts/wiki/Manual-Rollout-with-Puppet) as well. If Nick wants to chime in to the patchset as well, that would be even better.
> 
> I have some questions about rollout:
> 
>     1. Where do we get the non-ops-provided secrets from? Are we depending on getting them from another set up mac-v3 box, or are they going into the secrets repo, or? Do we know that another puppet run won't overwrite them? (Just double checking on the last)

At the moment I'm assuming we copy them from other nodes - I don't know the authoritative store for them. We need confirmation that puppet won't continue to run. 

>     2. How much testing have we done? Ideally we don't fully roll out until we know the various use cases are tested. We can roll out to a subset and let them run and monitor, or we can actively test, e.g. by calling [`create_test_workdir`](https://github.com/mozilla-releng/scriptworker-scripts/wiki/Testing-iscript#directory-tree) and manually running iscript, for a notarized dmg and geckodriver signing.

tb-mac-v3-signing2 and mac-v3-signing8 were both puppetised with this PR, and seem to be working (modulo the puppet daemon being restarted, which we presumably don't want). The manual test should also be fine.

>     3. Ideally we have a rollback plan if things break. That may be https://github.com/mozilla-releng/scriptworker-scripts/wiki/manual-rollout plus a copy of the pre-puppetized configs and secrets, from an unpuppetized box or a backup directory.

`rsync` from a different box would probably be simplest

modules/signing_worker/files/requirements.txt
```diff
@@ -0,0 +1,5 @@
+-e git+https://github.com/mozilla-releng/scriptworker-scripts#egg=mozbuild&subdirectory=vendored/mozbuild
+-e git+https://github.com/mozilla-releng/scriptworker-scripts#egg=scriptworker_client&subdirectory=scriptworker_client
+-e git+https://github.com/mozilla-releng/scriptworker-scripts#egg=iscript&subdirectory=iscript
```
nthomas-mozilla: Now that the mono-repo is populated with the other workers there are `production` and `production-<worker>` branches for them, while we're on master for iscript and deps. While we don't need to merge iscript to a branch to deploy I'm wondering if we can be consistent somehow ? Maybe a `production-iscript` branch that we use here, and some docs which point out how iscript differs from GCP-workers ?

modules/signing_worker/manifests/settings.pp
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class signing_worker::settings {
+
+    $virtualenv_dir = '/Users/cltbld/virtualenvs'
+    $user = 'cltbld'
+    $group = 'staff'
```
nthomas-mozilla: Is this a leftover ? I couldn't find anywhere it's referenced, and `virtualenv_dir` looks like it wouldn't work in the dep case.

modules/signing_worker/manifests/system_user.pp
```diff
@@ -1,18 +1,18 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-class roles_profiles::profiles::cltbld_user {
-
+class signing_worker::system_user (
```
nthomas-mozilla: Uh, lots of other roles_profiles use `include ::roles_profiles::profiles::cltbld_user`, so won't this file move break them ?

modules/signing_worker/manifests/base.pp
```diff
@@ -0,0 +1,173 @@
...
+        default: {
+            file { "${certs_dir}/widevine_prod.crt":
+                owner   => $signing_worker::user,
+                group   => $signing_worker::group,
+                content => base64('decode', lookup('signing_keys.widevine_prod_crt')),
+            }
+            file { "${certs_dir}/nightly-signing.keychain":
+                owner   => $signing_worker::user,
+                group   => $signing_worker::group,
+                content => base64('decode', lookup('signing_keys.nightly_signing_keychain')),
```
nthomas-mozilla: IIUC we've removed these keychains from the vault, so should this code me removed/commented out ?

modules/signing_worker/files/requirements.txt
srfraser: I'm happy with whatever branch I'm pointed to, so I'm assuming this is more a question for @escapewindow

modules/signing_worker/manifests/settings.pp
srfraser: I think it is, yes. I'll remove it.

modules/signing_worker/manifests/system_user.pp
srfraser: hmm, it didn't flag up any errors when running it. Odd. It seems it's being used for multiple purposes now, such as running a generic worker. What's the intended purpose of this user? Perhaps we should rename the one that's not for scriptworkers.

modules/signing_worker/files/requirements.txt
escapewindow: I don't love the `production*` branches and plan on obsoleting them (by porting scriptworker-scripts to taskgraph + CoT, and using relpro instead of force-pushes to ship docker images). However, these requirements.txt lines are candidates for future fixes as well: once we have wheel generation in the monorepo, we should use those. So I'm -0 on this :)

nthomas-mozilla: OK, lets leave it as is.

modules/signing_worker/manifests/system_user.pp
nthomas-mozilla: There might only be errors when those other roles attempt to include cltbld_user. Worst case you can restore the original file, right ?

nthomas-mozilla:
tb-mac-v3-signing2 somehow had a blank `worker_type` in scriptworker.yaml, when I attempted to start it up today.

srfraser:
> tb-mac-v3-signing2 somehow had a blank `worker_type` in scriptworker.yaml, when I attempted to start it up today.

I left that in a working state, so I don't know what changed there. Is puppet running an agent?

nthomas-mozilla:
It appears not - `/var/log/puppet/*` haven't changed since Oct 9, which corresponds to the 27 day uptime, and `/Library/LaunchDaemons/com.mozilla.atboot_puppet.plist` is the only puppet launcher I see. It looks like `org.mozilla.scriptworker.cltbld.plist` would keep trying to relaunch the scriptworker, so removing the worker type may have been a work around to keep it offline.

escapewindow:
@dividehex we keep losing our sudoers changes (to give cltbld passwordless sudo). Do you know if that's intentional? Is there a good way to prevent us from losing those?

escapewindow:
We may need to either update the `scriptworker.yaml` files to have

```
taskcluster_root_url: https://firefox-ci-tc.services.mozilla.com/
```

and/or update the default in scriptworker itself, and update the version of scriptworker. (This is for Saturday's cutover.)

modules/python/manifests/virtualenv.pp
```diff
@@ -0,0 +1,142 @@
...
+  $group            = 'root',
+  $mode             = '0755',
+  $proxy            = false,
+  $environment      = [],
+  $path             = [ '/bin', '/usr/bin', '/usr/sbin', '/usr/local/bin' ],
+  $cwd              = undef,
+  $timeout          = 1800,
+  $pip_args         = '',
+  $extra_pip_args   = '',
+  $virtualenv       = undef
```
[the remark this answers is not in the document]
srfraser: This file is imported from a module, I'd prefer not too drift too far from upstream unless it's a hard blocker.

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -31,16 +31,52 @@
 
             # we can add worker setup here like in gecko_t_osx_1014_generic_worker.pp
 
+            class { 'roles_profiles::profiles::logging':
```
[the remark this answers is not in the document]
srfraser: Should they not? I was following puppet's documentation. I can rework it to follow local style guides

modules/scriptworker_prereqs/manifests/init.pp
```diff
@@ -0,0 +1,47 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class scriptworker_prereqs {
+    contain packages::python3_s3
+    file { '/tools/python3':
+        ensure  => 'link',
+        target  => '/usr/local/bin/python3',
+        require => Class['packages::python3_s3'],
+    }
+    file { '/builds':
```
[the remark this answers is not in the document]
srfraser: Hm, so we have a different role for firefox, dep and thunderbird mac signing - but it uses the same profiles? Or refactor all of that out?

modules/roles_profiles/manifests/profiles/scriptworker_user.pp
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class roles_profiles::profiles::scriptworker_user {
+    case $::hostname {
+        /^dep-mac-v3-signing\d+/: {
+            class { 'signing_worker::system_user':
```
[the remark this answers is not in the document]
srfraser: It's worked on the test nodes (and I lifted the user creation from the existing code). You mentioned there were some oddities around creating system users that you'd implemented workarounds for, would that be covered if I used resources?

modules/roles_profiles/manifests/profiles/notarization_users.pp
```diff
@@ -0,0 +1,54 @@
...
+                $content = $::hostname? {
+                  /^(tb-)?mac-v3-signing\d+/ =>
+                        "cltbld ALL=(${user}) NOPASSWD: ALL\n",
+                  /^dep-mac-v3-signing\d+/ =>
+                        "${user} ALL=(root) NOPASSWD: /usr/bin/pkgbuild\n",
+                  default => fail('No matching hostname'),
+                }
+                sudo::customfile { "sudo_to_${user}":
+                  content => $content;
+                }
```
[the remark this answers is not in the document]
srfraser: I don't think we have a case where we want to give these users full admin - there's a workaround in system_user.pp at the moment due to the way sudo is being called, but that should hopefully change soon.

modules/roles_profiles/manifests/profiles/notarization_users.pp
```diff
@@ -0,0 +1,54 @@
...
+                    lookup('user_groups.dep_notarization', Array, undef, []),
+                default => [],
+            }
+            realize(Users::Single_user[$users])
+
+            # Monkey patching directoryservice.rb in order to create users also breaks group merging
+            # So we directly add the user to the group(s)
+            $users.each |String $user| {
+                exec { "${user}_developer_group":
+                    command => "/usr/bin/dscl . -append /Groups/admin GroupMembership ${user}",
```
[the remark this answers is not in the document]
srfraser: Good catch, need to double check with @nthomas

modules/signing_worker/manifests/system_user.pp
```diff
@@ -0,0 +1,66 @@
...
+                command => "/usr/bin/dscl . -append /Groups/_developer GroupMembership ${user}",
+                unless  => "/usr/bin/groups ${user} | /usr/bin/grep -q -w _developer",
+                require => User[$user],
+            }
+
+            # Set user to autologin
+            class { 'macos_utils::autologin_user':
+                user       => $user,
+                kcpassword => $kcpassword,
+            }
```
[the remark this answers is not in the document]
srfraser: I understood it was necessary for the scriptworker to function, which presumably means we need to test & rework how we have multiple scriptworkers operating on the same node - @escapewindow ?

modules/signing_worker/manifests/system_user.pp
```diff
@@ -0,0 +1,66 @@
...
+            sudo::custom { "allow_${user}_reboot":
+                user    => $user,
+                command => '/sbin/reboot',
+            }
+
+            # Consider removing this if/when 'sudo su' is no longer used.
+            sudo::custom { "allow_${user}_all":
+                user    => $user,
+                command => 'ALL',
+            }
```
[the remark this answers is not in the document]
srfraser: Hmm, which also means that the lines we're getting overwritten on the test nodes (see slack, and comments above) also need reworking. How should we best combine rules that say 'allow reboot as root, and access to the following other users' into one line?

modules/python/manifests/virtualenv.pp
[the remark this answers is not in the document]
dividehex: It isn't a hard blocker but if you're going to pull it from upstream then don't be afraid to own it downstream. 😸  Puppet data type declaration is just best practice, not syntax.

modules/signing_worker/manifests/base.pp
```diff
@@ -0,0 +1,173 @@
...
+        owner   =>  $signing_worker::user,
+        group   =>  $signing_worker::group,
+    }
+
+    vcsrepo { $widevine_clone_dir:
+      ensure   => present,
+      provider => git,
+      source   => "https://${widevine_user}:${widevine_key}@github.com/mozilla-services/widevine",
+    }
+    ->file { 'Remove widevine directory':
```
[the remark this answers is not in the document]
srfraser: Is there a local style guide I can refer to for this sort of thing? It seemed to be a standard puppet feature.

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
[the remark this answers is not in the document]
dividehex: I wouldn't rework it since the logging profile is used in  a lot of other roles.  It is on my list of things to fix and shouldn't be addressed in this scope of work.

modules/roles_profiles/manifests/profiles/notarization_users.pp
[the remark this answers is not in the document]
srfraser: Oops, wrong user. @nthomas-mozilla

modules/roles_profiles/manifests/profiles/notarization_users.pp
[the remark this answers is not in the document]
dividehex: What is the purpose of `cltbld ALL=(${user}) NOPASSWD: ALL\n`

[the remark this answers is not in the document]
srfraser: It's there while we wait for the change away from `sudo su` to `sudo -u` - this is the workaround I mention.

modules/roles_profiles/manifests/profiles/scriptworker_user.pp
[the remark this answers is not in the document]
dividehex: Fact is a classes should only be declared once if using resource-like declarations.  eg `class { 'signing_worker::system_user':}` because you aren't instantiating multiple instances of the class, only overriding a single instance with each declaration of that class at compile time.  Depending on puppets order of execution the results are unreliable.  And this isn't the behavior you want.  You want to instantiate these code blocks as multiple objects with different user parameters in the same catalog at the same time, therefore, you should user the defined resource block instead.  A defined resource block can instantiate multiple times because each has its own unique separate namespace.

See: [https://puppet.com/docs/puppet/latest/lang_classes.html#section-x54-1hk-xhb](https://puppet.com/docs/puppet/latest/lang_classes.html#section-x54-1hk-xhb)

Also, if this is being done elsewhere in the code, let me know so I can fix it.  I went looking but couldn't find it.

modules/roles_profiles/manifests/profiles/notarization_users.pp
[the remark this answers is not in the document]
escapewindow: I tried changing from `sudo su` to `sudo -u` and notarization died. We seem to be blocked on each other -- we have to make the sudoers change and the iscript change at the same time if we insist on switching to `sudo -u`.

modules/signing_worker/manifests/system_user.pp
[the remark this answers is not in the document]
escapewindow: If I understand the question, we do need a separate user per scriptworker daemon due to user-level XCode locks and keychain config, but that user doesn't need to be auto-logged-in to the console.

modules/signing_worker/manifests/system_user.pp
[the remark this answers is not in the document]
escapewindow: If need be, we could rename the scriptworker prod user to `cltsign`, which a) matches what we had on the signing servers and linux signing scriptworkers, and b) doesn't have a conflicting line in sudoers. Only mentioning this in case it helps work around the problem.

modules/roles_profiles/manifests/profiles/notarization_users.pp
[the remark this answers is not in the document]
nthomas-mozilla: I think the important thing is that notarization users (which are going to talk to apple) have read access to the zip file they need to upload, and the upload tool (`xcrun altool`). It's likely that making them an Admin was the quick way to do this when we were standing this up manually. I don't know what the developer group is, but it might solve the 2nd issue.

modules/roles_profiles/manifests/profiles/notarization_users.pp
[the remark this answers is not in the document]
nthomas-mozilla: Somewhat related - we don't need to call sudo pkgbuild from cltbld anymore, it went away as a ridealong in https://github.com/mozilla-releng/scriptworker-scripts/pull/21.

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -31,16 +31,97 @@
...
+
+            $scriptworker_users = lookup("scriptworker_users.${role}")
+
+            $scriptworker_users.each |String $user, Hash $user_data| {
+                signing_worker { "signing_worker_${user}":
+                    role                => $role,
+                    user                => $user,
+                    password            => lookup("${user}_user.password"),
+                    salt                => lookup("${user}_user.salt"),
+                    iterations          => lookup("${user}_user.iterations"),
```
dividehex: This is perfectly fine.  No need to merge hashes as discussed in slack.  Every time this resource is instantiated over the iteration, you have the keys (role and username) to lookup the secrets under a different but similar data structure in vault.yaml.  The lookup would look like:
```
password => lookup("scriptworker_secrets.${role}.${user}.password").
salt  => lookup("scriptworker_secrets.${role}.${user}.salt").
iterations => lookup("scriptworker_secrets.${role}.${user}.iterations").
```

and the corresponding secrets nested hash would look like this:
```
scriptworker_secrets:
  dep:
    depbld1:
		password:
		salt:
		iteration:
    depbld1:
		password:
		salt:
		iteration:
    tbbld:
		password:
		salt:
		iteration:
  tb-prod:
    cltbld:
		password:
		salt:
		iteration:
  ff-prod:
    cltbld:
		password:
		salt:
		iteration:
```

dividehex: Another reason not to attempt merging hashes while in hiera is, Vault (real vault, not vault.yaml) will be under a unique prefix namespace to signify the lookup is reaching into vault and not the in-tree yaml files.  A vault lookup will look like this:

` password => lookup("vault_secrets.scriptworker_secrets.${role}.${user}.password").`

srfraser: This is not the section from yesterday's discussion about merging, it's the worker and role config from lines 99 and 100. These are used to fill out templates, which have a variable number of sections in depending on the role the node is performing. See https://github.com/mozilla-platform-ops/ronin_puppet/pull/73/files/c5e4b0374a5531c3238613c2c89b48bf731401e7#diff-01a7e7a65760ec28c6ead2e535a11db6R17 for its current use.

modules/signing_worker/manifests/system_user.pp
```diff
@@ -0,0 +1,60 @@
...
+define signing_worker::system_user (
+    String $user,
+    String $password,
+    String $salt,
+    String $iterations,
+) {
+    case $::operatingsystem {
+        'Darwin': {
+
+            # Create the cltbld user
```
davehouse: old/invalid comment?

modules/signing_worker/manifests/system_user.pp
```diff
@@ -0,0 +1,60 @@
...
+            users::single_user { $user:
+                # Bug 1122875 - cltld needs to be in this group for debug tests
```
davehouse: old/invalid (no group specified)

PR #75, opened 2019-08-13
1561956_generic-worker_15

dragoscrisan:
Update generic-worker to version 16.5.2 multiuser on OSX Mojave


modules/generic_worker/templates/generic-worker.plist.erb
```diff
@@ -12,7 +12,7 @@
         <string>/usr/local/bin/run-generic-worker.sh</string>
         <string>run</string>
         <string>--config</string>
-        <string>/etc/generic-worker.config</string>
+        <string>/etc/generic-worker/config</string>
```
davehouse: Does this need to go back to a dot to match the reference in  modules/generic_worker/manifests/multiuser.pp (https://github.com/mozilla-platform-ops/ronin_puppet/pull/75/files#diff-fb9e83eb4fe169b0d24255d1dd4a5180R82)

davehouse: nm, you pushed a new commit switching to using this /etc/generic-worker directory. 👍

modules/generic_worker/manifests/multiuser.pp
```diff
@@ -0,0 +1,107 @@
...
+                    group   => $::root_group;
+
+                '/usr/local/bin/run-generic-worker.sh':
+                    ensure  => present,
+                    content => template('generic_worker/run-generic-worker.sh.erb'),
+                    mode    => '0755',
+                    owner   => $::root_user,
+                    group   => $::root_group;
+
+                '/etc/generic-worker.config':
```
davehouse: I'll double-check later that you changed this reference too :)

modules/generic_worker/templates/generic-worker.config.erb
```diff
@@ -30,16 +26,16 @@
...
   "tasksDir": "<%= @task_dir %>",
   "workerGroup": "<%= @worker_group %>",
   "workerId": "<%= @hostname %>",
   "workerType": "<%= @worker_type %>",
+  "wstAudience": "taskcluster-net",
+  "wstServerURL": "https://websocktunnel.tasks.build",
   "workerTypeMetadata": {
     "machine-setup": {
-      "config": "https://github.com/mozilla-platform-ops/ronin_puppet/raw/master/modules/generic_worker/templates/generic-worker.config.erb",
+      "config": "https://github.com/mozilla-releng/build-puppet/raw/master/modules/generic_worker/templates/generic-worker.config.erb",
```
dividehex: Why change this back to the old config in build-puppet? it should be point to here.

modules/generic_worker/manifests/multiuser.pp
dividehex: You changed the references to the config file to `/etc/generic-worker/config` but this is still putting the config in `/etc/generic-worker.config`

modules/generic_worker/templates/generic-worker.plist.erb
dividehex: If you are going put the config in `/etc/generic-worker/config`, you should also create the dir `/etc/generic-worker/`

dragoscrisan:
Looks like pre-commit hooks fixed a end line error on modules/fluentd/templates/application_default_credentials.json.erb. This file is not part of my changes

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -13,6 +13,17 @@ rm -f tasks-resolved-count.txt
 export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
 source /usr/local/share/generic-worker/bugzilla-utils.sh
 
+. /etc/rc.common
```
davehouse: I don't think you need to do this network check because puppet already waits for network before running and creating the semaphore file that triggers this.

Also, we aren't controlling rc.common. So this *could* do anything.

modules/generic_worker/templates/generic-worker.plist.erb
davehouse: thx. I see the working dir paths are all matching now for multiuser. It is a little confusing to have the config in different places with similar names.

modules/generic_worker/templates/generic-worker.config.multiuser.erb
```diff
@@ -0,0 +1,42 @@
...
+  "certificate": "",
+  "checkForNewDeploymentEverySecs": 0,
+  "cleanUpTaskDirs": true,
+  "clientId": "<%= @taskcluster_client_id %>",
+  "deploymentId": "",
+  "downloadsDir": "<%= @downloads_dir %>",
+  "idleTimeoutSecs": 345600,
+  "instanceId": "",
+  "instanceType": "",
+  "livelogExecutable": "livelog",
```
davehouse: Is livelog deprecated? Maybe in a follow-up change we can remove this and the livelog install since we have wst* set and working.

modules/generic_worker/templates/run-generic-worker.sh.erb
davehouse: Please remove this network check block.

davehouse: Thanks! that was causing me stress ;)

modules/talos/manifests/init.pp
```diff
@@ -23,7 +23,7 @@
                     ensure  => directory,
                     owner   => $user,
                     group   => 'staff',
-                    mode    => '0755',
+                    mode    => '0777',
```
davehouse: do these need to be wide open, or can they be made only writable by the generic-worker users' group or main user?

modules/generic_worker/templates/run-generic-worker.sh.multiuser.erb
```diff
@@ -0,0 +1,75 @@
...
+# exits, this script will reboot the machine. Note, in the worker config
+# (/etc/generic-worker.config) we set numberOfTasksToRun to 1 so that the worker
+# will exit after running a single task. This script is then responsible for
+# rebooting the machine.
+
+# If this file exists when the worker starts up, and its content is equal
+# to numberOfTasksToRun, the worker won't exit and machine won't reboot.
+rm -f tasks-resolved-count.txt
+
+export SHELL=/bin/bash
```
petemoore: You can delete this line, it won't set `SHELL` for tasks. See e.g. [this task](https://tools.taskcluster.net/groups/BH_28jjfRN-5ZA-T1t8JUQ/tasks/BH_28jjfRN-5ZA-T1t8JUQ/runs/0/logs/public%2Flogs%2Flive.log). It runs against production, using this version of `run-generic-worker.sh.multiuser.erb` but SHELL is not set in python process. `SHELL` is a [special bash variable](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-SHELL), that evaluates to the user's shell if the environment variable `SHELL` is not set. Since the spawned process for the task does not inherit the exported environment variables of the generic-worker process, `SHELL` will not be set, and so a task that executes a bash shell will have the bash variable `SHELL` set, but any processes it runs will not have the `SHELL` environment set.

modules/generic_worker/templates/generic-worker.config.multiuser.erb
petemoore: The livelog executable is still needed when using websocktunnel, but the livelog secrets are not needed, as access to the livelog stream (provided by the livelog executable) is provided via the tunnel, rather than a direct client-to-worker ssl connection that was previously used. The former secrets enabled the service to be hosted over https on the worker directly, and the livelog secret was used in tandem with the stateless dns server in order to provide hashed hostnames for the workers.

modules/generic_worker/templates/generic-worker.plist.multiuser.erb
```diff
@@ -0,0 +1,42 @@
...
+<%# file, You can obtain one at http://mozilla.org/MPL/2.0/. -%>
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>LSEnvironment</key>
+    <dict>
+        <key>SHELL</key>
+        <string>/bin/bash</string>
+    </dict>
```
petemoore: This SHELL section should be removed.

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker_staging.pp
```diff
@@ -60,14 +58,17 @@
                 quarantine_client_id      => $quarantine_client_id,
                 quarantine_access_token   => $quarantine_access_token,
                 bugzilla_api_key          => $bugzilla_api_key,
-                generic_worker_version    => 'v13.0.3',
-                generic_worker_sha256     => '6e5c1543fb3c333ca783d0a5c4e557b2b5438aada4bc23dc02402682ae4e245e',
+                task_dir                  => '/Users',
+                generic_worker_version    => 'v15.1.4',
```
petemoore: 16.4.0 is the current latest version.

modules/generic_worker/manifests/multiuser.pp
```diff
@@ -0,0 +1,114 @@
...
+        livelog_version           => $livelog_version,
+        livelog_sha256            => $livelog_sha256
+    }
+
+    class { 'generic_worker::control_bug':
+        user_homedir     => $user_homedir,
+        bugzilla_api_key => $bugzilla_api_key,
+    }
+
+    $livelog_certificate = "${user_homedir}/livelog.crt"
```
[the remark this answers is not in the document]
davehouse: we have this for the single-user/old-versions generic-worker until they're all moved to multi-user/15+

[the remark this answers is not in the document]
dragoscrisan: multiuser class is used only for define generic-worker new version. Old version is defined in init.pp

modules/generic_worker/templates/generic-worker.config.multiuser.erb
davehouse: thx

davehouse: for the explanation 👍  I won't expect to remove the livelog binary now

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker_staging.pp
```diff
@@ -51,23 +48,25 @@
...
-                livelog_secret            => $livelog_secret,
                 worker_group              => $worker_group,
                 worker_type               => $worker_type,
                 quarantine_client_id      => $quarantine_client_id,
                 quarantine_access_token   => $quarantine_access_token,
                 bugzilla_api_key          => $bugzilla_api_key,
-                generic_worker_version    => 'v13.0.3',
-                generic_worker_sha256     => '6e5c1543fb3c333ca783d0a5c4e557b2b5438aada4bc23dc02402682ae4e245e',
+                task_dir                  => '/Users',
+                generic_worker_version    => 'v16.4.0',
```
petemoore: I don't understand why you haven't hit [bug 1592233](https://bugzilla.mozilla.org/show_bug.cgi?id=1592233).

modules/generic_worker/templates/generic-worker.config.multiuser.erb
```diff
@@ -0,0 +1,42 @@
...
+  "livelogPUTPort": 60022,
+  "numberOfTasksToRun": 1,
```
petemoore: We should change this to `0`. A value of `1` implies you only intend to run one task on the worker, so no new task user will be created for the subsequent task. Setting to 0 implies you intend to run an unlimited number of tasks on this worker.

modules/generic_worker/templates/run-generic-worker.sh.multiuser.erb
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+# This is a wrapper script for calling generic worker, such that when the worker
+# exits, this script will reboot the machine. Note, in the worker config
+# (/etc/generic-worker.config) we set numberOfTasksToRun to 1 so that the worker
+# will exit after running a single task. This script is then responsible for
+# rebooting the machine.
```
petemoore: This whole paragraph can be removed. It is correct for the simple engine, but for multiuser engine, `numberOfTasksToRun` needs to be `0` so that task users are created for future tasks.

modules/generic_worker/templates/run-generic-worker.sh.multiuser.erb
```diff
@@ -0,0 +1,69 @@
...
+
+# If this file exists when the worker starts up, and its content is equal
+# to numberOfTasksToRun, the worker won't exit and machine won't reboot.
+rm -f tasks-resolved-count.txt
```
petemoore: This `rm` and the comment above it is also not needed for multiuser when `numberOfTasksToRun` is `0`.

modules/generic_worker/templates/generic-worker.config.multiuser.erb
```diff
@@ -12,7 +12,7 @@
   "instanceType": "",
   "livelogExecutable": "livelog",
   "livelogPUTPort": 60022,
-  "numberOfTasksToRun": 1,
+  "numberOfTasksToRun": 0,
```
davehouse: 😢

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker_multiuser.pp
```diff
@@ -46,8 +46,8 @@
                 quarantine_access_token   => $quarantine_access_token,
                 bugzilla_api_key          => $bugzilla_api_key,
                 task_dir                  => '/Users',
-                generic_worker_version    => 'v16.4.0',
-                generic_worker_sha256     => 'eaf18131ff7f7e28810554719c71e5b658bec374d03ca469c5b9af272920e331',
+                generic_worker_version    => 'v16.5.2',
+                generic_worker_sha256     => '7bd47da57aae65f120d89e8d70fb0a1f66762945994e0909d31eac6d63122046',
```
petemoore: https://tools.taskcluster.net/groups/FEKcrpGcQ4uRXVJ6-cl9PQ/tasks/JCtMJWbbRz2xVWgkpbdWjg/runs/0/logs/public%2Flogs%2Flive_backing.log#L14 is from a couple of hours ago and shows `t-mojave-r7-244` is in the staging pool (`releng-hardware/gecko-t-osx-1014-staging`) but running generic-worker 15.1.4.

dragoscrisan:
> ).

Removed

modules/generic_worker/manifests/multiuser.pp
```diff
@@ -0,0 +1,129 @@
...
+            require => Class['packages::generic_worker'];
+    }
+
+    case $::operatingsystem {
+        'Darwin': {
+
+            $reboot_command = '/usr/bin/sudo /sbin/reboot'
+
+            file {
+                # Ensure old plist and config files doesn’t exists, and deletes yhem, if necessary.
```
petemoore:
```suggestion
                # Ensure old plist and config files doesn’t exists, and deletes them, if necessary.
```

PR #77, opened 2019-08-21
1561956 generic worker 15

dragoscrisan:
Upgrade generic worker to v15.1.4 - staging workers


dividehex:
I've rebased staging to master so both branches are current.  Please start a new PR against staging with just your changes.

PR #79, opened 2019-08-21
Bug  1559442

markcor:
Upgrade generic-worker to 15.1.0  for datacenter Win 10 workers.


markcor:
@davehouse, yes, and think that would be awesome.

PR #82, opened 2019-08-22
1571949 - Port nrpe install to ronin puppet

dragoscrisan:
Port nrpe install to ronin puppet


nthomas-mozilla:
FWIW, we'd like to monitor mac-v3-signing boxes like a scriptworker, eg [this balrog one](https://nagios1.private.releng.mdc1.mozilla.com/releng-mdc1/cgi-bin/status.cgi?navbarsearch=1&host=balrog), except no puppet freshness check.

So that leaves us with 
* PING (??)
* pending scriptworker tasks (check_pending_scriptworker_tasks)
  * if this pulls in a lot of dependencies then we can look at implementing it elsewhere
  * see https://github.com/mozilla-releng/build-puppet/blob/master/modules/scriptworker/manifests/nagios.pp#L10
* scriptworker log age (check_scriptworker_file_age, [derived from check_file_age](https://github.com/mozilla-releng/build-puppet/blob/master/modules/scriptworker/manifests/nagios.pp#L13))
* disk (check_nrpe_disk)
* load (check_nrpe_load)
* ntp peer (check_ntp_peer)
* ntp time (check_ntp_time)
* scriptworker procs (check_nrpe_procs_regex)

Does that help set the scope of the request better ?

nthomas-mozilla:
I see PING comes from a clustercheck on the nagios server instead.

dividehex:
Closing in favor of Duplicate #135

PR #87, opened 2019-08-29
configure telegraf on devicepool hosts

aerickson:


modules/bitbar_devicepool/files/telegraf.conf
```diff
@@ -0,0 +1,52 @@
...
+  flush_interval = "1m"
+  flush_jitter = "0s"
+  precision = ""
+  debug = false
+  quiet = false
+  ## Override default hostname, if empty use os.Hostname()
+  hostname = ""
+
+[[outputs.influxdb]]
+  # these values are replaced at converge time by bootstrap script
```
davehouse: why replace it then instead of using hiera variables?

aerickson: The values are stored in encrypted in SOPS and replaced with the real credentials after convergence.

https://github.com/mozilla-platform-ops/android-tools/blob/master/devicepool_bootstrap_and_setup/devicepool_update.sh#L56

PR #92, opened 2019-09-09
Bug1570767 - Add Chrome install to Win 10 hw workers

markcor:


modules/win_packages/manifests/chrome.pp
```diff
@@ -0,0 +1,40 @@
...
+    $chrome_key       = "HKLM\\SOFTWARE\\Policies\\Google\\Chrome"
+    $chrome_reg_value = [
+                        "${chrome_key}\\MetricsReportingEnabled",
+                        "${chrome_key}\\SafeBrowsingExtendedReportingEnabled",
+                        "${chrome_key}\\ChromeCleanupReportingEnabled"
+                        ]
+    # Pulling directly from source to ensure latest release
+    # Existing installs will update automaticly
+    file { $installer_path:
+        source => "https://dl.google.com/tag/s/dl/chrome/install/${installer}",
```
dividehex: I think we should be more explicit and point to our s3 bucket.   I was under the impression that we wanted to lock chrome to a certain version so we can explicitly upgrade when we need to.

modules/roles_profiles/manifests/profiles/chrome.pp
```diff
@@ -0,0 +1,17 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
dividehex: Not sure you need an profile here but if you are, call it google_chrome and not just chrome

modules/win_packages/manifests/chrome.pp
markcor: @dividehex  In https://bugzilla.mozilla.org/show_bug.cgi?id=1570767#c6 I was told we wanted to keep it at the latest. I am reluctant to put in an s3 bucket because of the unknown behavior of an old version being installed and then having it jump multiple versions.

modules/roles_profiles/manifests/profiles/chrome.pp
markcor: @dividehex I renamed the profile. I want to keep this as profile, so it is explicit in the role if Chrome is installed or not.

modules/win_packages/manifests/chrome.pp
markcor: @dividehex I moved the source to S3 and noted the version in the file tags. If it causes an issue in the future we can just update the package. 

I am hesitant to put  the version into Puppet since it will reflect what is on the actual worker.

dividehex: I understand the need to "keep it at latest" but I don't think it is a good idea to just let the application update itself at random across test workers.  It is highly likely we will end up with tests running on multiple version of chrome due to the fact some machines may or may not have upgraded yet.  It also runs the risk of a self update failing an we don't know about it.  Imagine gw forcing a reboot in the middle of an chrome update or chrome updating itself in the middle of the test.  Or you might end up with the chrome restarting half way through the test.  If those are acceptable events then so be it.  Otherwise, we should have a way to prevent self updating and controlling chrome upgrading through puppet deployment or at least control when self updating happens.

PR #98, opened 2019-09-16
Bug1580502 zstd macos

davehouse:
macos cli binaries for zstandard 1.3.8 (matching the version in the python zstandard library we'll also install, https://github.com/indygreg/python-zstandard/blob/0.11/zstd/zstd.h#L72)

### DIFF
--- a/modules/packages/manifests/zstandard.pp
+++ b/modules/packages/manifests/zstandard.pp
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 class packages::zstandard (
-    Pattern[/^v\d+\.\d+\.\d+$/] $version,
+    Pattern[/^\d+\.\d+\.\d+$/] $version,
 ) {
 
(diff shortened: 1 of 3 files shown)


modules/packages/manifests/zstandard.pp
```diff
@@ -0,0 +1,16 @@
...
+class packages::zstandard (
+    Pattern[/^v\d+\.\d+\.\d+$/] $zstandard_version,
+    String                      $zstandard_sha256,
+) {
+
+    packages::macos_package_from_s3 { "zstd-${zstandard_version}.dmg":
+        private             => false,
+        os_version_specific => false,
+        type                => 'dmg',
+        checksum            => $zstandard_sha256,
```
dividehex: dmg(w/pkg inside) don't utilize hashes.  Only the bin type uses them to ensure the file isn't downloaded on every puppet run.

davehouse: thx. I removed the sha entries

PR #99, opened 2019-09-16
break-fix typo

davehouse:


dividehex:
@davehouse Thanks for catching and fixing

PR #100, opened 2019-09-16
Bug1580502 zstd macos python

davehouse:
this will use the wheel available on pypi: https://pypi.org/project/zstandard/#files

I looked at putting the wheel into s3, but I am using pypi to keep it simpler and because we do the same for virtualenv/others


dividehex:
fix lint before merging

PR #106, opened 2019-09-24
[Staging] [WIP] Bug1572749 macos to papertrail

davehouse:
Send macos logs to papertrail:
1. Translate macos log entries to syslog:
   a. Rewrite log entry severity (ex: "Default" -> "notice", "Error" -> "err")
   b. Capture program name
1. Add output for fluentd on macos to send to remote syslog (log-aggregators).
1. Reduce log captures to Default/notice or more severe (excluding Info)


davehouse:
the generic-worker multiuser setup is still on staging; I'll send this to master instead

PR #111, opened 2019-09-30
remote_syslog not remote-syslog

davehouse:
I left it with the wrong remote_syslog package. 
remote_syslog takes severity and program/etc from the event, and uses the current plugin interface. 
remote-syslog does not take the syslog tags from the event and uses the deprecated fluent plugin interface.


davehouse:
I manually installed this gem on the workers friday evening (when I found no logs in papertrail after merging the change to copy logs to papertrail). So this change will control/manage the already installed plugin.

davehouse:
When this is merged, I'll manually remove the one that was installed mistakenly ("remote-syslog").

PR #114, opened 2019-10-04
Add independent module settings and root owners/groups vars

dividehex:
This may seem a little redundant but it is in an effort to make modules more self contained.  Also, global vars suck especially with puppet apply.


davehouse:
I understand the idea for self-contained modules, but i think adding manifests increases the complexity and I think globals are a good place for some facts we share across modules (just like $::operatingsystem).
"especially with puppet apply" -- do the globals break something?

dividehex:
> I understand the idea for self-contained modules, but i think adding manifests increases the complexity and I think globals are a good place for some facts we share across modules (just like $::operatingsystem).
> "especially with puppet apply" -- do the globals break something?

First, this is all driven by the work I'm doing in macstadium with vault/bolt_provisioning.

Mainly, globals completely break the ability to load directly into a class eg `-e 'include roles_profiles::roles::somerole'`  In the _new_ bolt provisioning, we don't want to generate nodes instead we want include classes directly without a site.pp or node scope.

In a puppet apply world, you shouldn't need a node scope at all so I want to get rid of the /manifests and simply load direct classes.  This PR is the first step. This also allows you to use hiera or consul as an ENC if we want to which I've been toying with as an idea.  You can also see that having a globals breaks kitchen tests as @aerickson noted in the repo [1].  And I'd like to follow @aerickson 's lead and write kitchen tests for other systems also.

I know there might be a better way to load a global like a shared module but it isn't like we are going to need one spot to change that code.  linux will always be `root:root` and macos/bsd will always be `root:wheel`.

If this _**REALLY**_ bothers you, we could do a shared module such as `include shared; $shared::root_user`.  I personal prefer settings.pp in each module.

[1] https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/manifests/site.pp#L5

dividehex:
Actually... I've decided to refactor this. 🏭  We can be explicit about file attributes without site/node scope global variables _AND_ without repeating ourselves.  I think the ::shared module might be a better idea.

PR #140, opened 2019-11-09
Bug1593727 - update TC root url and rotate secrets 

markcor:


modules/win_generic_worker/templates/hw-generic-worker.config.epp
```diff
@@ -26,7 +26,7 @@
   "queueBaseURL": "",
   "region": "",
   "requiredDiskSpaceMegabytes": 10240,
-  "rootURL": "https://taskcluster.net",
+  "rootURL": "<%= $win_generic_worker::hw_config::taskcluster_root %>",
```
davehouse: nice. (it is hard-coded for macos too) this looks like it will be easier to manage

PR #159, opened 2019-11-26
Bug 1595971 - clean up registry profile list  

markcor:
This PR addresses a few different items.

Primarily adds a schedule task to Windows static workers that will remove profile list registry entries from deleted task users. I am dding this as a separate schedule task because I want to keep this separate from the maintain system task.  I would like to keep the maintain system task generic as possible and use it across all Windows platforms. 

Cleans up the static worker gw run bat be moving the deletion of  tasks-resolved-count.txt to only occur after a gw panic and adds a log message for when poissbly corrupted files are being removed post gw panic. 

Update the wstaudience and wstserverurl, so that live logging will work with firefox-ci-tc. This alos moves the value of these into Hiera. 

General clean up: 
Remove old gw profiles that are no longer in use. 

Remove the roles_profiles::profiles::windows_rotate_secrets from static worker roles. This one is not effectively doing anything at this point. I am leaving the class in place, but will need to have a file comparison add before it is used int he future.

Test try push:   
https://treeherder.mozilla.org/#/jobs?repo=try&revision=24d4d29e455f68984faa65a75472cd95cced1089
https://treeherder.mozilla.org/#/jobs?repo=try&revision=31be3f91b134aaa2de907424c86a9b9c8493de23


modules/win_generic_worker/templates/run-hw-generic-worker-and-reboot.bat.epp
```diff
@@ -53,11 +70,13 @@ shutdown /r /t 0 /f /c "Generic worker has reached idle timeout; rebooting" >> <
...
 :exit_69
+echo "Generic-worker panic! Issue with enviroment or worker bug. (exit code %GW_EXIT_CODE%)" >> <%= $win_generic_worker::generic_worker_dir %>\generic-worker-wrapper.log
+echo "Removing possibly corrupted files." >> <%= $win_generic_worker::generic_worker_dir %>\generic-worker-wrapper.log
+del /Q /F <%= $win_generic_worker::generic_worker_dir %>\tasks-resolved-count.txt
 del /Q /F <%= $win_generic_worker::generic_worker_dir %>\file-caches.json
 del /Q /F <%= $win_generic_worker::generic_worker_dir %>\directory-caches.json
 del /Q /F <%= $win_generic_worker::generic_worker_dir %>\next-task-user.json
 del /Q /F <%= $win_generic_worker::generic_worker_config %> >> <%= $win_generic_worker::generic_worker_dir %>\generic-worker-wrapper.log
-echo "Generic-worker panic! Issue with enviroment or worker bug. (exit code %GW_EXIT_CODE%)" >> <%= $win_generic_worker::generic_worker_dir %>\generic-worker-wrapper.log
 Goto ErrorReboot
```
davehouse: on posix, we've seen workers reboot "too quickly" and end up thrashing in the 69 state when a taskcluster service was down.
Is that a risk on Windows? On posix, we ended up just stopping the worker (so we have to manually go un-block them since either they're broken or a TC service is down). But I think we'll change it to a 1-2min delay and then reboot so they don't repeat their failure quickly.

markcor: On Windows we were seeing one of the json files or a single text file  end up with a data that was causing a Go error. I wonder if generic worker was writing to those files when the reboot happened which could possibly cause an issue with the data being written.  Like generic worker exits but the service is still writing out to files when the wrapper script calls a reboot.

PR #160, opened 2019-11-27
Splits out mac signing roles

dividehex:


dividehex:
DS now splits these into 3 different groups with unique puppet_roles.

PR #164, opened 2019-12-05
Scriptworker atboot

srfraser:


srfraser:
I can't add a reviewer - @dividehex ?

srfraser:
@davehouse

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -10,27 +10,6 @@
             $worker_type  = 'mac-v3-signing'
             $worker_group = regsubst($facts['networking']['fqdn'], '.*\.releng\.(.+)\.mozilla\..*', '\1')
 
-            # TODO: mac-v3-signing should run puppet more than just at boot.
-            # This needs a puppet::periodic class for running puppet on a cron schedule
-            class { 'puppet::atboot':
```
davehouse: Why are we removing the puppet run at boot?

srfraser: Not all of our secrets are suitable for the vault, so we are manually adding some, and puppet would overwrite them. `vcsrepo` was also causing trouble trying to clone into a directory that already existed, causing errors for Puppet when run multiple times - this is implementing both of the options (Don't run at boot, check for existence) that were discussed with Jake. None of the options provided in this class appear to do anything, so it didn't seem like much of a loss.

davehouse: thx!

PR #165, opened 2019-12-12
linux 18.04 base, MAAS, and gecko_t_linux_talos roles

aerickson:
## overview

- adds the following role_profiles
  - linux 1804 base
  - gecko_t_linux_talos
  - maas region and maas rack
- sets up test-kitchen testing for the above

---

## notes on gecko_t_linux_talos

Ports `toplevel::worker::releng::generic_worker::test::gpu` from build_puppet to ronin_puppet.

Top level definition in this repo is at `modules/roles_profiles/manifests/roles/gecko_t_linux_talos.pp`.

### SKIPPED

- graphics drivers (intel)
  - intel p580 drivers should be included
  - see https://bugzilla.mozilla.org/show_bug.cgi?id=1424465#c3 about dhouse's previous issues
- alsa sound driver
  - use pulse now?

### secrets that need setting

```
            $user = lookup('linux_vnc_user')
            $user_homedir = lookup('linux_vnc_user_homedir')
            $group = lookup('linux_vnc_group')
            $password = lookup('linux_vnc_password')
```

### how to test

see README.md in this commit


davehouse:
thank you! I have not used kitchen so this is a huge help to have it ready.

aerickson:
For sure. :) I'll update the README.md with the howto I have above.

Should I just continue adding everything that the gecko-t-linux-talos/t-linux64-ms hosts in build-puppet have?

davehouse:
Yes, I'm good with that.

aerickson:
Talked with Jake. This isn't 100% yet, but he thought we should land to avoid bit rot.

aerickson:
rebased

modules/linux_vnc/manifests/init.pp
```diff
@@ -0,0 +1,44 @@
...
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class linux_vnc {
+  # include config
+  # include users::builder
+  include linux_packages::vnc_server
+  include roles_profiles::profiles::cltbld_user
```
dividehex, line 11: do not include profiles inside component modules

modules/linux_vnc/manifests/init.pp
```diff
@@ -0,0 +1,44 @@
...
+
+  if (lookup('cltbld_user.vnc_password') == '') {
+    fail('No VNC password set')
+  }
```
dividehex, line 15: hiera lookups should only be done within profiles and then passed as args to the class

modules/linux_packages/manifests/python3_zstandard.pp
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class linux_packages::python3_zstandard {
+    require linux_packages::py3
+
+    # for the first run, this tries to run pip because pip3 didn't exist
+    # at the start at the start of the run. ugh. puppet bug in 5.5.1.
```
dividehex, line 9: Is the still a bug in puppet 6?  We should be using the latest puppet.

aerickson: It was always working in 6. Not sure why I thought we were locked at 5 (maybe the servers I looked at were using it or ???). It was awhile ago when I wrote that.

Will change.

modules/linux_vnc/manifests/init.pp
aerickson: This instance has been fixed.

modules/linux_vnc/manifests/init.pp
aerickson: Fixed all instances.

modules/linux_vnc/manifests/init.pp
aerickson: All instances fixed.

PR #167, opened 2020-01-06
mac signing worker monitoring

davehouse:
@srfraser the telegraf module is ready for configuring on the signing scriptworkers. If this looks good to you, I'll test it on one of the mac-v3 workers (or you can if you want) and then we can merge it.

### DIFF
--- a/modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
+++ b/modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
@@ -127,6 +127,54 @@ class roles_profiles::profiles::mac_v3_signing {
                     ed_key_filename     => $user_data['ed_key_filename'],
                 }
             }
+
+            class { 'telegraf':
+                global_tags  => {
+                    workerType    => $worker_type,
+                    workerGroup   => $worker_group,
+                    provisionerId => 'scriptworker-prov-v1',
+                    role          => $role,
+                    workerId      => $facts['networking']['hostname'],
+                },
+                agent_params => {
+                    interval          => '300s',
+                    round_interval    => true,
+                    collection_jitter => '0s',
+                    precision         => 's',
+                },
+                inputs       => {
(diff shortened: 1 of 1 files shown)


modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -127,6 +127,54 @@
...
+                        interval => '60s',
+                        exe      => '/builds/scriptworker/bin/scriptworker',
+                    },
+                    procstat2   => {
+                        plugin_type => 'procstat',
+                        interval    => '60s',
+                        pattern     => 'tools/release/signing/signing-server.py',
+                    },
+                    puppetagent => {
+                        location => '/opt/puppetlabs/puppet/cache/state/last_run_summary.yaml',
```
srfraser: Haven't we disabled puppet agent runs in #164 ? I can't merge that other PR, which is why it's not active yet.

davehouse: I'll merge that PR #164 now, and then we can see the real diff for scriptworkers.

srfraser: I think this code block is all new, so the diff won't change. Puppet agent shouldn't be running, in any case, but I don't know how this particular check works - is it looking for the age of the last run?

davehouse: Here's the readme and code for the pupppetagent plugin: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/puppetagent
It won't alert itself on anything, but reads in most of the values from the yaml file. From that, we can see the age of the last run and success/failure.

davehouse:
@srfraser I'm merging this telegraf setup for the signing scriptworkers in master now (ready for when you want to apply it on the machines)

PR #170, opened 2020-01-08
Bug1587085 macos chrome update

davehouse:


davehouse:
sparky confirmed with a test pool that the update is good

PR #175, opened 2020-01-15
Bug1609467 disable puppet atboot

davehouse:


davehouse:
tested on a test/try worker and confirmed the launchdaemon plist is removed

PR #184, opened 2020-03-11
[Bug1563799] remove homebrew

dividehex:
This PR is a WIP and picks up on the work dhouse already did. (thanks dave!)

This code works but needs more testing.


davehouse:
I ran this on another worker this morning and hit this error with the homebrew uninstaller:
```
Error: Warning: The Ruby Homebrew uninstaller is now deprecated and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall.sh)"

/bin/bash: line 184: brew: command not found
/bin/bash: line 189: HOME: unbound variable
```

modules/macos_utils/manifests/uninstall_homebrew.pp
```diff
@@ -0,0 +1,13 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class macos_utils::uninstall_homebrew {
+
+    exec { 'run_homebrew_uninstaller':
+        path      => '/usr/bin:/usr/sbin:/bin',
+        command   => 'ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)"',
+        onlyif    => 'test -e /usr/local/Homebrew',
```
davehouse, line 10: I adjusted like this for my testing:
https://github.com/mozilla-platform-ops/ronin_puppet/blob/f81ed37e0e2b9a6994eb751a6b1994973a9c6e24/modules/macos_utils/manifests/uninstall_homebrew.pp

I had to change to checking for the brew binary because the uninstaller leaves /usr/local/Homebrew (maybe just the new one does this).
maybe we need to delete the directory after the uninstall?

dividehex: The uninstall also checks for the brew bin so testing for that is the right option

dividehex:
I've tested this on both a fresh ds install and on an existing staging host.  I think it is about time to merge this so I would like a thorough review before merging it.

PR #188, opened 2020-04-13
1561956 generic worker 15 multiuser

davehouse:


davehouse:
closed to silence the spam from travis ci

PR #190, opened 2020-04-24
bug 1549046: allow macos user to change desktop background

dividehex:


dividehex:
If you inspect the bin, it changes the `locked` keys value to `false`, allowing the macos user to change the desktop background.
```
              <key>locked</key>
              <false/>
```

PR #198, opened 2020-06-11
restore macos cltbld autologin

davehouse:
revert part of 7d687c42


davehouse:
something different is wrong. (the autologin moved into the cltbld user class)

PR #199, opened 2020-06-17
Bug1639352 signing bootstrap

davehouse:
@escapewindow this moves signing over to py3.8.3 (and fixes the virtualenv dependency for a new reimaged machine to run puppet)


modules/scriptworker_prereqs/manifests/init.pp
```diff
@@ -27,14 +30,6 @@
             returns => [1];
     }
 
-    # Install certifi's set of CAs to override the system set
-    exec {
-        'install_python_certs':
-            command => "'/Applications/Python 3.7/Install Certificates.command'",
```
davehouse, line 33: This certs setup is done in packages::python3

PR #200, opened 2020-06-17
bug 1645229 symlink /Library under /usr for macos hg

davehouse:


modules/packages/manifests/mercurial.pp
```diff
@@ -14,4 +14,16 @@
         os_version_specific => false,
         type                => 'pkg',
     }
+
+    # pkg installs /usr/local/bin/hg
+    # which looks for the mercurial packages in:
+    # libdir = '../../Library/Python/2.7/site-packages/'
+    # but installs them in:
+    # /Library/Python/2.7/site-packages/
```
davehouse, line 22: This was found as a problem by a task that ran hg in an environment without the python2 site-packages in the path. This caused hg to fail when trying to import hgdemandimport. Other tasks/things appear to not be failing because they have the python2/3 paths in the env path.

PR #176, opened 2020-01-22
[bug 1619377] - super rough sketch of notarization poller, for feedback

escapewindow:
@srfraser , @davehouse , I'm looking to add a notarization-poller worker per prod/tb-prod mac:

- this should run on prod/tb-prod, not dep
- we can install notarization-poller in the same venv as iscript
- we need a separate user/pass to run notarization-poller
- we need the notarization user/pass secrets from the `script_config.yaml`, but we'll also add a unique clientId/accessToken for the worker
- we'll need a different config file (`poller.yaml`) and a separate launchd entry / wrapper
- I'm thinking about running these in a `poller/` subdirectory

I see a few ways to do this:

- put a complex `if` block into `signing_worker/init.pp`, to conditionally set up the poller, or
- create a new .pp file that we conditionally call from `signing_worker/init.pp`; this new `.pp` file contains all poller logic
- something else? New module, and the prod/tb-prod machines also use this module?

Do you recommend a particular direction here at the outset?

PR #203, opened 2020-06-23
Notarization poller

escapewindow:
See #176 for history.

This patchset:
- adds the notarization poller
  - adds a poller user and config
- updates the `supported_behaviors`
- sets the workerGroup to mdc1 or mdc2 as appropriate
- adds a `scriptworker_version` and `scriptworker_scripts_revision`, so in the future we can bump these without SSHing in
  - updates requirements.txt.erb with those
  - removes an errant requirements.txt file

I had to update the vault with the firefoxci cluster accessTokens and the poller secrets.

For the future:
- fix dep. It looks like puppetizing the dep instances is broken on master, and continues to be broken with this patchset. I think the Duplicate errors we're getting are largely related to running `init.pp` 3 times, rather than using an `instance.pp`.
- if we can get `pip install` to run on `requirements.txt` changes, and the scriptworker and notarization_poller daemons to reload on python or config file change, then we can upgrade packages without sshing into the boxes. This would be a big win.
- if we have improved our secrets management/security story, then we may be able to put our final 4 secrets into the vault. This would mean we could image and puppetize the signing boxes, and go live without human intervention.


escapewindow:
@dividehex @davehouse how does this look? (Attempts to fix dep are in https://github.com/escapewindow/ronin_puppet/commits/akitest ; I think we have to start over with an instance.pp to really fix.)

escapewindow:
> r+
> This looks good. and +1 that notarization_poller handles logging itself

Thanks!

> Futures:
> 
>     1. Is there urgency to fix dep? Or can you make a bug for relops to fix it in Q3/4? building off [escapewindow/ronin_puppet@notarization-poller...escapewindow:akitest](https://github.com/escapewindow/ronin_puppet/compare/notarization-poller...escapewindow:akitest) or making an instance.pp

Non urgent, but it would be good for the dep boxes to be cattle, not pets. Avoiding ssh as part of our maintenance is preferred. I'll file a bug, thank you!

>     2. Would updates to requirements.txt packages require puppet to run on a schedule, or would the updates only be for new imaged workers? (at-boot puppet is still disabled for signing)

This would mean we need to run puppet on a schedule. Doable? This is a reason why we need to ssh in to maintain the mac signers, which we want to avoid. I'm hoping we can take some of the prior art from puppetagain in terms of restarting services on change.

I tested -- rerunning puppet looks like it doesn't blow away the `certs/` keychains, keys, and certs. We would need some way to update puppet and the vault periodically.

>     3. @dividehex do you want for us to plan to move signing over to using the vault service?

If so, we can potentially reimage and get a working mac signer at the end, but we do have to have some security guarantees for the secrets. If not, we can live with the one-time setup hit.

escapewindow:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1648845 for dep.

PR #221, opened 2020-08-31
Bug1661748 setup loaner pool

davehouse:
for mac loaners:
1. open firewall for vpn ssh/vnc
2. put into single-worker workerType pools


dividehex (changes requested):
Passing parameters to a profile breaks the roles/profile model. I've been meaning go through and clean up where it exists else where in the code base.  I think we can either create a new user just for loaners that has and use that instead of cltbld or just add the screenshare and ssh groups to the cltbld user across the board.  IIRC, that was how it was originally setup in the old puppet.

For `su` access, I would suggest a dedicated profile like profiles::cltbld_loaner

PR #224, opened 2020-09-25
Bug 1662170 - Install Pyhton3 certifi 

markcor:


davehouse (approved):
+1 we had to do this on the mac minis for python3 also

PR #226, opened 2020-10-12
stop pinning hg ssl fingerprint

davehouse:


davehouse:
@cgsheeh what do you think: Could we stop pinning the hg ssl fingerprint on the CI test workers?

aerickson (approved):
r+

PR #237, opened 2020-12-11
Update bolt config to comply with bolt v2.37.0

dividehex:
This makes the config compliant with bolt 2.37.  The `host-key-check` is moved to the users config under `~/.puppetlabs/etc/bolt/bolt-defaults.yaml` since it shouldn't be assumed to be project wide.


davehouse (approved):
lgtm

PR #241, opened 2020-12-22
Integrate multi-user generic worker

dividehex:
This PR is an attempt to integrate some outstanding branches that are being used in production.  It primarily merges the multiuser generic worker bits.


davehouse:
r+ for the latest merge. a diff confirms the changes are testing updates, the bolt inventory, linux1804, and the addition of psutil.

modules/packages/manifests/xcode_cmd_line_tools.pp
```diff
@@ -4,7 +4,7 @@
 
 class packages::xcode_cmd_line_tools {
 
-    packages::macos_package_from_s3 { 'Command_Line_Tools_macOS_10.14_for_Xcode_10.1.dmg':
+    packages::macos_package_from_s3 { 'Command_Line_Tools_for_Xcode_12.2.dmg':
```
dividehex, line 7: This might be big breaking change.  In the future (soon) each role will specify which package version it expects.  In the meantime, either base version on OS major version or test if the new xcode cli breaks on Mojave or Catalina.

modules/roles_profiles/manifests/profiles/cltbld_user.pp
```diff
@@ -2,8 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-class roles_profiles::profiles::cltbld_user {
-
+class roles_profiles::profiles::cltbld_user (
```
dividehex, line 5: As a rule of thumb, we should not be passing parameters to any profiles.  I intend to fix every profile that does this currently.  Although, since provides defaults that align with existing settings, I'm inclined to let this apply and fix it later when we fully implement role hiera data.

modules/roles_profiles/manifests/profiles/gecko_1_b_osx_1015_generic_worker.pp
```diff
@@ -0,0 +1,160 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class roles_profiles::profiles::gecko_1_b_osx_1015_generic_worker (
+    String $worker_type = 'gecko-1-b-osx-1015',
```
dividehex, line 6: The `$worker_type` seems unique to this profile anyway.  Let's just move this down a couple lines and remove it from the parameters.

modules/roles_profiles/manifests/profiles/gecko_3_b_osx_1015_generic_worker.pp
```diff
@@ -0,0 +1,160 @@
...
+            #include packages::google_chrome
+            #file { '/var/opt/generic-worker':
+            #    ensure => 'directory',
+            #    mode   => '0755',
+            #}
+
+            contain packages::nodejs
+            #contain packages::wget
+            contain packages::tooltool
+            file { '/tools/tooltool.py':
```
dividehex, line 115: For both Catalina and Bigsur, we need to stop setting up `/tools` and `/builds` since both OSes implement a RO `/` filesystem.  I know you can remount it RW but we really shouldn't.  I'm pretty sure Apple will prevent that in the future.

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker_staging.pp
```diff
@@ -4,7 +4,9 @@
 
 class roles_profiles::profiles::gecko_t_osx_1014_generic_worker_staging {
 
-    require roles_profiles::profiles::cltbld_user
+    class { 'roles_profiles::profiles::cltbld_user':
+        autologin => false,
```
dividehex, line 8: If a worker is running mutliuser, I don't think we even need the `cltbld` user.  Generic worker will create its own users and set autologin accordingly.

modules/talos/manifests/init.pp
```diff
@@ -11,7 +11,7 @@
             include httpd
             include packages::java_developer_package_for_osx
             include packages::xcode_cmd_line_tools
-            require dirs::builds
+            #require dirs::builds
```
dividehex, line 14: Why is this commented out?

dividehex:
@davehouse Can you have a look at my comments above?  My primary concern is anything that might break our largest MacOS workerpool `gecko-t-osx-1014`.

modules/mercurial/manifests/cacert.pp
```diff
@@ -4,7 +4,7 @@
 
 class mercurial::cacert {
 
-    include dirs::builds
+    #include dirs::builds
```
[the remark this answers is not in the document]
davehouse: puppet cannot create the /builds path and when it is an existing hardlink (from synthetic.conf), puppet was still trying to change the permissions on it and failing.

modules/packages/manifests/generic_worker.pp
```diff
@@ -7,11 +7,11 @@
...
     Pattern[/^v\d+\.\d+\.\d+$/] $taskcluster_proxy_version,
     String                      $taskcluster_proxy_sha256,
-    Pattern[/^v\d+\.\d+\.\d+$/] $quarantine_worker_version,
-    String                      $quarantine_worker_sha256,
+    Pattern[/^v\d+\.\d+\.\d+$/] $livelog_version,
+    String                      $livelog_sha256,
 ) {
 
-    packages::macos_package_from_s3 { "generic-worker-darwin-amd64-${generic_worker_version}":
+    packages::macos_package_from_s3 { "generic-worker-multiuser-darwin-amd64-${generic_worker_version}":
```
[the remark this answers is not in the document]
davehouse: sgtm

modules/packages/manifests/xcode_cmd_line_tools.pp
davehouse: agreed. this is needed for the builders, but for testers I think keeping the old commandlinetools is safer

modules/roles_profiles/manifests/profiles/cltbld_user.pp
davehouse: I'm okay with removing this. It was leftover from loaner testing and isn't used by the builders.

modules/roles_profiles/manifests/profiles/gecko_1_b_osx_1015_generic_worker.pp
davehouse: np, this isn't necessary

modules/roles_profiles/manifests/profiles/gecko_3_b_osx_1015_generic_worker.pp
davehouse: I'm okay with moving these elsewhere, but isn't the synthetic.conf setup is simpler to then match the expectation of /tools etc on linux?

modules/roles_profiles/manifests/profiles/gecko_t_osx_1014_generic_worker_staging.pp
davehouse: sounds fine to me. This was to ensure removing the cltbld user from the autologin, but is not necessary with generic worker setting up the autologin task user.

modules/talos/manifests/init.pp
davehouse: same as the other, puppet tries to change permission on the synthetic.conf hard links

modules/generic_worker/templates/generic-worker.plist.erb
```diff
@@ -27,5 +32,9 @@
             <true/>
         </dict>
     </dict>
+    <key>StandardErrorPath</key>
```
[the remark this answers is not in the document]
davehouse: +1. This was to get the worker-runner logs in case worker-runner fails. So once run-generic-worker is spawned, this log includes a duplicate of those logs

modules/roles_profiles/manifests/profiles/gecko_3_b_osx_1015_generic_worker.pp
dividehex: I think we need to change that expectation.  It has always been poor form to assume you have write access to the root filesystem.

dividehex:
@davehouse I intend to merge this tomorrow 03/02/21. It is pushed to staging right now.  Can you do one last review?

modules/roles_profiles/manifests/profiles/gecko_3_b_osx_1015_generic_worker.pp
davehouse: Do you want to add a synthetic.conf to the bootstrap so that it can be used in bigsur also until the tasks are changed to not require a /tools ? Or do you want to remove this now, and that forces the change as part of greening-up the tasks to run on R8?

davehouse (approved):
r+ and here are unnecessary, but nice to have, tweaks:
https://github.com/dividehex/ronin_puppet/pull/1
https://github.com/dividehex/ronin_puppet/pull/2
https://github.com/dividehex/ronin_puppet/pull/3

dividehex:
At this time, I don't want to use synthetic.conf.  I think we should just move to change the tests to find tooltool in more conforming PATH location like `/usr/local/bin`.  We have already done this on the M1 Bigsur and r8 catalina workers.

Going forward we should work to minimize the need to alter the underlying OS and keep as many of the security features in place as possible.  eg keep SIP enabled, gatekeeper enabled, secure boot enabled, keep RO filesystem enabled, etc, etc.

PR #242, opened 2020-12-22
Define more r8 minis in inventory assigned puppet roles

dividehex:
Here we call out how r8 minis are currently being utilized per pool in the bolt inventory files


aerickson (approved):
syntax lgtm

PR #243, opened 2020-12-22
Simple bolt deploy plan to exec puppet on a remote target

dividehex:
Adds a simple bolt plan to deploy puppet to a remote target.  Also, requires pre-commit puppet validate to ignore the `.pp` files in modules plans dir since it can't differentiate between puppet and bolt plan files.


aerickson (approved):
lgtm

PR #244, opened 2020-12-22
Introduce the use of role data in hiera

dividehex:
This PR introduces the use of role data which will significantly simplify workers how are given custom data and should allow for future (soon) clean up of the multiple role/profile matches.  And all the duplicate code that comes with it.  This also gives a nice single file view of a roles data profile and will also allow for easier version bumping on various thing on a per-role basis.


aerickson (approved):
lgtm

davehouse (approved):

markcor (approved):

PR #245, opened 2020-12-23
First round of package classes versioning

dividehex:
There should be a `version` parameter on any package class where there might be a version to specify.  This PR is the first effort to make that happen.  It won't cover everything but slowly chip away at this PR by PR.


aerickson (approved):
lgtm

PR #247, opened 2020-12-23
Add a logrotation resource for macos

dividehex:
This PR adds a simple macos resource for rotating log files using the MacOS newsyslog.  Which is essentially a log rotation cron job set to evaluate a `run-parts` style dir of conf files that describe log files and how to rotate them.


aerickson (approved):
lgtm

PR #250, opened 2021-01-11
fix mac dep signing puppetization

bhearsum:
This isn't done yet, but it's a start. Most of the issues I've found so far are around not having unique names for some resources.

### DIFF
--- a/modules/signing_worker/manifests/init.pp
+++ b/modules/signing_worker/manifests/init.pp
@@ -101,6 +101,7 @@ define signing_worker (
       force    => true,
       user     => $user,
       group    => $group,
+      require  => File[$scriptworker_base],
     }
     # This has credentials in it. Clean up.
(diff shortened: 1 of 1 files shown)


bhearsum:
I think this is ready for review. Besides the name issues, there was a couple of missing dependencies, permissions problems, typedef issues, and a bad check for `poller_config`. @escapewindow - feel free to take a look for functionality, but I think you said @davehouse or @dividehex will need to do the for realz review.

I've tested this against dep so far by hacking mac-v3-signing13 to install that. I still need to test with prod and tb roles before this can be merged.

escapewindow (dismissed):
LGTM, thank you!
And yes, we should double-check that we don't regress prod with this. It looks like it's probably safe, though.

dividehex (dismissed):
👍 lgtm

dividehex (approved):

PR #254, opened 2021-01-15
bitbar-devicepool: disable bclary

aerickson:
Disable by setting login shell to /usr/sbin/nologin. Verify in serverspec.

Also:
- minor gem and kitchen tweaks
- circleci runs precommit --verbose to show times


modules/bitbar_devicepool/manifests/base.pp
```diff
@@ -38,11 +38,16 @@
     ensure => 'present',
   }
 
+  # disable login for departed users
```
davehouse, line 41: good idea to ensure users cannot login and are gone. (considering that just removing user creation could miss something)

davehouse (approved):

PR #256, opened 2021-01-15
Fix dep scriptworker configuration

bhearsum:
Without this, only one can be running at a time, which is problematic for machines that multiple instances.

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -50,6 +50,7 @@ scriptworker_users:
         - mac_sign
         - mac_sign_and_pkg
       dmg_prefix: "tb"
+      worker_type_prefix: "tb-"
       worker_id_suffix: "tb"
       cot_product: "thunderbird"
       widevine_filename: "widevine_dep.crt"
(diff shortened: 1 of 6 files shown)


davehouse (dismissed):

bhearsum:
Sorry for all of the churn here - there was more issues than I expected! This is now working well on dep-mac-v3-signing1, and ready for review.

bhearsum:
@escapewindow, this can probably use your review too.

davehouse (approved):

escapewindow (approved):
The `tb_` config living inside the `ff` config is weird, but I'm happy to have it working.
Thanks!

bhearsum:
> The `tb_` config living inside the `ff` config is weird, but I'm happy to have it working.
> Thanks!

yeah, it's not ideal, but i didn't want to reorganize all of those configs as part of this work. even if the tb dep workers go away we may want similar for adhoc? at some point we should detach role from products that have instances.

bhearsum:
I discovered a lingering issue today -- some of the dep parameters weren't correctly set. The latest push should fix that. We also need to update the vault in a couple of ways:
* `scriptworker_secrets.tb-prod.taskcluster_client_id` -> `scriptworker_secrets.tb-prod.tb_taskcluster_client_id`
* `scriptworker_secrets.tb-prod.taskcluster_access_token` -> `scriptworker_secrets.tb-prod.tb_taskcluster_access_token`
* Add `scriptworker_secrets.dep.tb_taskcluster_client_id` and `scriptworker_secrets.dep.tb_taskcluster_access_token` (which I'll need to send you.

davehouse (approved):

PR #258, opened 2021-01-21
Pin signing_worker dependencies

bhearsum:
This gets all of the Python dependencies (both scriptworker and other things that we use directly, as well as their dependencies) pinned. I've had to change the way signing workers are installed and set-up a bit to make this work. With this applied, the install goes like this:
1) Create the virtualenv
2) Install the pinned dependencies into the virtualenv
3) Clone scriptworker, et. al via a pinned revision and install those into the virtualenv

The dependencies are version and hash pinned in the `files` directory of the `signing_worker` module. Because they can be configured on a per workerType basis, we have 3 copies of the requirements. I've also added a helper script to make it easier to update them. The way it works is to define or reference all of the things that scriptworker, et. al depend on - using the pinned revision of those repos from `common.yaml` whereever possible.

Whenever we take updates to scriptworker or scriptworker-scripts, the process should be:
* Land the scriptworker and/or scriptworker-scripts changes
* Update the revision(s) in `common.yaml`
* Update the requirement file(s) with the helper script
* Create a PR for this repo with the new revisions + requirements

@escapewindow - I also realized that our discussion the other day about needing to pin the deps on Mac is less crucial than I thought. That's only going to become an issue if we have a package that is _only_ required on Mac. Binary packages are fine, as `pip-compile` adds all hashes (even ones for the platforms you're not running it on) to the generated requirements file). Given that, I'm not going to worry about that part for the time being. We'll deal with it if it comes up in the future.

I've tested this on an ff-prod worker so far, but it _should_ work for the others. I'll do some more testing on other worker types before landing assuming this looks OK.


bhearsum:
I should also note that right now this is pointing at my `scriptworker` repo, until https://github.com/mozilla-releng/scriptworker/pull/479 is merged.

bhearsum:
@dhouse / @dividehex - can one of you take a look at this (looks like I can't ask for formal review)?
@escapewindow - do you want to review this too?

modules/signing_worker/files/update-requirements.sh
```diff
@@ -0,0 +1,59 @@
...
+# Where possible we reference their requirements.in files. Where not possible, or they
+# don't exist, we list them directly (which means manual updates if they change).
+#
+# The packages listed above are installed directly from source clones, so they do not need
+# to go in the requirements file.
+cat <<EOF > requirements.in
+-r https://raw.githubusercontent.com/mozilla-releng/scriptworker-scripts/$scriptworker_scripts_revision/scriptworker_client/requirements/base.in
+-r https://raw.githubusercontent.com/mozilla-releng/scriptworker-scripts/$scriptworker_scripts_revision/iscript/requirements/base.in
+-r https://raw.githubusercontent.com/mozilla-releng/scriptworker-scripts/$scriptworker_scripts_revision/notarization_poller/requirements/base.in
+-r https://raw.githubusercontent.com/mozilla-releng/scriptworker/$scriptworker_revision/requirements.in
```
escapewindow, line 47: requirements.txt ?

modules/signing_worker/files/update-requirements.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# mozbuild dependencies
+jsmin
+mozfile
+
+# widevine dependencies
+cryptography
+macholib
```
escapewindow, line 55: Ah, thank you. I was wondering how we were dealing with these. Since these change rarely, this works.

modules/signing_worker/manifests/init.pp
```diff
@@ -126,23 +224,16 @@
...
+        user        => $user,
+        group       => $group,
+        refreshonly => true,
+        subscribe   => [Exec["clone widevine ${scriptworker_base}"], Python::Virtualenv["signingworker_${user}"]],
+        require     => [Exec["clone widevine ${scriptworker_base}"], Python::Virtualenv["signingworker_${user}"]],
     }
+
     # XXX once we:
     #     - get the virtualenv to re-run pip on requirements.txt change,
     #     - get the scriptworker and poller to restart on config or python
```
escapewindow, line 239: It looks like we still need this at some point, yes? Essentially, this PR will install a new version of scriptworker onto a running machine, but the old version of scriptworker will remain running until we ssh in and manually kick it. Accurate?

(I'm ok with that as a follow-up; just making sure we track it if it's still needed work.)

escapewindow (approved):
Thanks! We should still get relops' review here :)

modules/signing_worker/files/update-requirements.sh
bhearsum: Yeah, it's not ideal, but I think it's a fine compromise for now.

modules/signing_worker/manifests/init.pp
bhearsum: Yes, that's right. I don't think it's safe to fix this part until we have secrets managed by Puppet.

davehouse (approved):

PR #260, opened 2021-01-25
1804: generic-worker idleTimeoutSecs should be much higher (4d)

aerickson:


davehouse (approved):
96h instead of 1.5? lgtm

PR #262, opened 2021-01-29
fix shellcheck on circleci

aerickson:
Fixes the current pre-commit failure on master (https://app.circleci.com/pipelines/github/mozilla-platform-ops/ronin_puppet).

- Switches to our fork of https://github.com/jumanjihouse/pre-commit-hooks that removes the additiionalDependency parameter to the pre-commit hook. It's implied/unecessary and causes pre-commit to explode on circleCI.
- Switches pre-commit to only run on the files changed between HEAD and master.
  - Runs much more quickly. 
  - More inline with the intent of pre-commit.

--

Research:

shellcheck-py recommendation: https://github.com/pre-commit/pre-commit/issues/1453. Exploded on circleCI (see 
838090d  below).

users of this hook complaining: https://github.com/jumanjihouse/pre-commit-hooks/issues/82. No solution found.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,9 +9,8 @@ jobs:
       - run: sudo apt-get update
       - run: sudo apt-get install shellcheck
       - run: pip3 install pre-commit
-      - run: shellcheck --version
-      - run: ls -la
-      - run: pre-commit run --all-files --verbose
+      #- run: pre-commit run --all-files --verbose
+      - run: pre-commit run --files `git diff --name-only HEAD HEAD~1`
   r10k_install:
     machine:
       image: ubuntu-2004:202010-01
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,9 +36,9 @@ workflows:
   ci_test:
     jobs:
       - pre_commit
-        #- r10k_install
-        #- kitchen_converge_and_verify:
-        # matrix:
-        #   parameters:
-        #     # "maas-region" and "maas-rack" removed as we're not using them currently
-        #     kitchen_target: ["bitbar", "linux"]
+      - r10k_install
+      - kitchen_converge_and_verify:
+          matrix:
+            parameters:
+              # "maas-region" and "maas-rack" removed as we're not using them currently
+              kitchen_target: ["bitbar", "linux"]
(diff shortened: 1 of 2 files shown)


davehouse (approved):
\o/

.circleci/config.yml
```diff
@@ -9,7 +9,8 @@ jobs:
       - run: sudo apt-get update
       - run: sudo apt-get install shellcheck
       - run: pip3 install pre-commit
-      - run: pre-commit run --all-files --verbose
+      #- run: pre-commit run --all-files --verbose
+      - run: pre-commit run --files `git diff --name-only HEAD master`
```
dividehex, line 13: Does this work instead?
`pre-commit run --from-ref origin/HEAD --to-ref HEAD`

It seems more inline with running in CI.

https://pre-commit.com/#usage-in-continuous-integration

aerickson: On circleCI, origin/HEAD should be identical master so that does the same thing I think (in a more elegant way). I missed those options. Will update.

aerickson:
There's a workaround to the issue in https://github.com/jumanjihouse/pre-commit-hooks/issues/82#issuecomment-770077155 that doesn't involve forking the repo (although I think we should control our pre-commit hooks and selectively merge anyways). Thoughts?

aerickson:
Hm, I think on CircleCI origin/HEAD could be incorrect in terms of a PR. That is origin is my fork, not the canonical source (this repo).

My origin/HEAD is out of date (not up to date with mozilla-platform-ops/ronin_puppet HEAD). We should add a remote pointing at `https://github.com/mozilla-platform-ops/ronin_puppet` and diff against that.

dividehex:
> 
> 
> There's a workaround to the issue in [jumanjihouse/pre-commit-hooks#82 (comment)](https://github.com/jumanjihouse/pre-commit-hooks/issues/82#issuecomment-770077155) that doesn't involve forking the repo (although I think we should control our pre-commit hooks and selectively merge anyways). Thoughts?

Yes! let's just add `additional_dependencies: []` and call this done.

dividehex (approved):
Thanks Andy! Let's merge this as is.

PR #265, opened 2021-02-05
update iscript+scriptworker for taskQueueId + openh264 zipfile

escapewindow:
@bhearsum


bhearsum (approved):
For future reference, you don't have to pick up dependency changes whenever you take scriptworker/iscript changes IMO. This is fine though.

davehouse (approved):

PR #266, opened 2021-02-09
bump scriptworker+scriptworker-scripts

escapewindow:
To:
- fix dep bustage around pkg signing.
- prepare for a future taskcluster version that introduces `projectId`

@bhearsum


escapewindow:
(I already rolled this out to dep boxes via `python setup.py develop`, but ideally we roll out the real fix soon.)

davehouse (approved):

PR #274, opened 2021-02-22
Bug1693164 fix mac mojave signer logging

davehouse:
* make generic-worker log tail optional (disabled for signers)


aerickson (approved):
lgtm

PR #281, opened 2021-03-09
talos 1804: install py3.9

aerickson:
During my testing for the 1804 migration, we missed some tier 3 jobs that aren't normally scheduled. They need  python 3.8 or greater.

Install python3.9 (sourced from the deadsnakes PPA). Python3.6 (provided by ubuntu) is still present.

It's 4.4 MB of packages.

Testing on 002.

### DIFF
--- a/modules/linux_packages/manifests/py3.pp
+++ b/modules/linux_packages/manifests/py3.pp
@@ -4,11 +4,42 @@
 
 class linux_packages::py3 {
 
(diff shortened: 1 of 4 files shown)


aerickson:
The packages installed fine on 002.

davehouse (approved):
Awesome. That's great+simple through s3!

PR #277, opened 2021-02-27
(wip) [RELENG-378] Periodic mac signer puppet

escapewindow:

PR #284, opened 2021-03-11
[RELENG-378] Periodic mac signer puppet

### LINKED ISSUE #267: run puppet on mac signers periodically
https://jira.mozilla.com/browse/RELENG-230

@dividehex @davehouse we're planning on running puppet on the mac signers periodically, so we can land dependency bumps and config changes and have them take effect without having to ssh in. This should reduce the security concern of sshing in and the maintenance concern of human error.

Thoughts on this? Should we just add a wrapper script that pulls the latest ronin_puppet and runs puppet, or is there a better way of doing this?

escapewindow:
Supersedes #277. Fixes #267.

This PR:
- adds a `puppet::periodic` module with a "lockfile" directory to avoid concurrent puppet runs,
- enables `puppet::periodic` in the mac signers,
- prevents the mac signers from starting up on first image/puppet (we need to populate the signing secrets, or we'll burn the tree),
- adds wrapper scripts to help with maintenance

### DIFF
--- a/modules/puppet/files/com.mozilla.periodic_puppet.plist
+++ b/modules/puppet/files/com.mozilla.periodic_puppet.plist
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+
+    <!-- This Source Code Form is subject to the terms of the Mozilla Public
+       - License, v. 2.0. If a copy of the MPL was not distributed with this
+       - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+                <key>LANG</key>
+                <string>en_US.UTF-8</string>
+                <key>PATH</key>
+                <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/puppetlabs/bin</string>
+        </dict>
+        <key>Label</key>
+        <string>com.mozilla.periodic_puppet</string>
+        <key>StartInterval</key>
+        <integer>900</integer>
+        <key>RunAtLoad</key>
+        <false/>
+        <key>ProgramArguments</key>
+        <array>
+                <string>/bin/bash</string>
+                <string>/usr/local/bin/periodic-puppet.sh</string>
+        </array>
+        <key>ServiceIPC</key>
+        <false/>
+        <key>StandardErrorPath</key>
+        <string>/var/log/puppet/puppet.err</string>
+        <key>StandardOutPath</key>
+        <string>/var/log/puppet/puppet.out</string>
+</dict>
+</plist>
--- a/modules/puppet/templates/puppet-darwin-periodic-puppet.sh.erb
+++ b/modules/puppet/templates/puppet-darwin-periodic-puppet.sh.erb
@@ -1,0 +1,24 @@
+#!/bin/bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# in the absence of `lockfile`, use a locking dir
+# https://unix.stackexchange.com/a/274
+LOCK_DIR="/tmp/.periodic-puppet"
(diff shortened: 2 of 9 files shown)


modules/puppet/files/com.mozilla.periodic_puppet.plist
```diff
@@ -0,0 +1,35 @@
...
+        <dict>
+                <key>LANG</key>
+                <string>en_US.UTF-8</string>
+                <key>PATH</key>
+                <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/puppetlabs/bin</string>
+        </dict>
+        <key>Label</key>
+        <string>com.mozilla.periodic_puppet</string>
+        <key>StartInterval</key>
+        <integer>900</integer>
```
escapewindow, line 20: Guessing 15 min is a decent number.

modules/puppet/files/com.mozilla.periodic_puppet.plist
```diff
@@ -0,0 +1,35 @@
...
+        <key>RunAtLoad</key>
+        <false/>
```
escapewindow, line 22: I'm thinking when we first load this, a puppet run is already happening, so let's not kick off another one.

modules/puppet/files/com.mozilla.periodic_puppet.plist
```diff
@@ -0,0 +1,35 @@
...
+        <key>ProgramArguments</key>
+        <array>
+                <string>/bin/bash</string>
+                <string>/usr/local/bin/periodic-puppet.sh</string>
```
escapewindow, line 26: Wrapper script to use a "lockfile" directory.

modules/puppet/templates/puppet-darwin-periodic-puppet.sh.erb
```diff
@@ -0,0 +1,24 @@
...
+    exit 1
+}
+
+function finish {
+    rm -rf $LOCK_DIR
+}
+
+mkdir "${LOCK_DIR}" || fail
+trap finish EXIT
+$PUPPET 2>&1 > $LOG_FILE
```
escapewindow, line 24: This means that `/var/log/puppet/puppet.out` doesn't get very much output. I can remove the redirects, or keep them (the /tmp log file only keeps the latest run).

modules/signing_worker/manifests/init.pp
```diff
@@ -71,16 +73,20 @@
 
     $required_directories = [
       $scriptworker_base,
-      "${scriptworker_base}/certs",
       "${scriptworker_base}/logs",
-      "${scriptworker_base}/artifact",
```
escapewindow, line 76: `artifact` is unused, and scriptworker creates the `artifacts` directory on start.

modules/signing_worker/manifests/init.pp
```diff
@@ -71,16 +73,20 @@
...
       ensure => 'directory',
       owner  =>  $user,
       group  =>  $group,
       mode   => '0750',
     }
+    file { "${scriptworker_base}/certs":
+      ensure => 'directory',
+      owner  =>  $user,
+      group  =>  $group,
+      mode   => '0700',
```
escapewindow, line 88: Since multiple scriptworker processes with different users may be on the box, I'm setting the yaml files and certs directories to user-only perms. (Not the base directory because `poller` is a subdirectory currently, and shares the virtualenv.)

dividehex (approved):
LGTM

modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -10,8 +10,6 @@
             $worker_type  = 'mac-v3-signing'
             $worker_group = regsubst($facts['networking']['fqdn'], '.*\.releng\.(.+)\.mozilla\..*', '\1')
 
-            include puppet::disable_atboot
```
bhearsum, line 13: Just for my own clarity: we're not using this because these are long running machines instead of ones that reboot per job?

I'm a bit surprised we couldn't have tweaked this module, since it's already a thing that knows how to run puppet, but I suspect you probably either tried it, or had good reasons not to. (I don't think it's worth doing at this point, certainly.)

bhearsum (approved):

escapewindow:
Thank you both! We'll need to create a `production-mac-signing` branch here (can we/should we limit who can push there?)

escapewindow:
> Thank you both! We'll need to create a `production-mac-signing` branch here (can we/should we limit who can push there?)

I was able to create the branch. We're probably just limited to whoever has write access to the repo currently.

PR #288, opened 2021-03-17
Bug1693164 papertrail macs direct

davehouse:
I've been testing this with t-mojave-r7-286(singleuser) and macmini-r8-139(catalina,multiuser).
I haven't tested this on an r8 with bigsur.


modules/macos_utils/manifests/set_hostname.pp
```diff
@@ -0,0 +1,15 @@
...
+
+    if $::operatingsystem == 'Darwin' {
+        exec { 'SetLocalHostName':
+            command => '/usr/sbin/scutil --set LocalHostName $(hostname|cut -d\. -f1)',
+            onlyif  => '/bin/test $(/usr/sbin/scutil --get LocalHostName) != "$(hostname|cut -d\. -f1)"',
+        }
+    } else {
+        fail("${module_name} does not support ${::operatingsystem}")
+    }
+}
```
dividehex, line 15: I don't think this should be set here.  Historically, we have set this on deployment with deploystudio.  In the r8 and m1 world, we are managing the LocalHostName in MDM.  So that would conflict.  If r7s do not have LocalHostName set correctly, then we should fix this in deploystudio and then do a bolt command to fix all the existing nodes.

modules/roles_profiles/manifests/profiles/logging.pp
```diff
@@ -55,16 +57,19 @@
             # https://bugzilla.mozilla.org/show_bug.cgi?id=1520947
         }
         'Darwin': {
+            include macos_utils::set_hostname
```
dividehex, line 60: If we were to set the hostname in puppet, we would definitely want to make it a profile to be called out in roles so it isn't buried randomly in the logging profile.  But either way, I don't think puppet should be touching the hostname/scutils.  It should be done once on deployment or actively managed in MDM.

dividehex (changes requested):

modules/macos_utils/manifests/set_hostname.pp
davehouse: Great! I didn't consider MDM and saw the install-created LocalHostName coming up in syslog for the R8's. I'll remove this.

aerickson (approved):
looks sane to me, not sure if you want to wait for dividehex's r+

dividehex (approved):
👍

PR #290, opened 2021-03-17
Fix periodic puppet

escapewindow:
When testing the periodic puppet rollout, I noticed that we reboot the mac 2x after imaging. This means that even though we don't load the `/Library/LaunchDaemons/*.plist` files during puppet time, we load them on boot, and we start up scriptworker before we populate the signing secrets.

This PR moves the `<%= scriptworker_base %>/.enabled` check into the poller+scriptworker wrapper scripts, rather than the launchd wrappers, so we don't start up scriptworker on boot before we're ready.

### DIFF
--- a/modules/signing_worker/templates/launchctl_wrapper.sh.erb
+++ b/modules/signing_worker/templates/launchctl_wrapper.sh.erb
@@ -1,20 +1,11 @@
 #!/bin/bash
 # Wrapper to [re]start a scriptworker
 
+set +e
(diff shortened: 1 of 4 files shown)


bhearsum (approved):
This looks ok to me. Just a sanity check though: is `enable_scriptworker.sh` meant to be called by hand? (I don't see anything Puppet-managed that calls it, but maybe I'm misreading)

escapewindow:
It is. Until the signing secrets are populated from vault, we run this by hand after we populate the signing secrets.

PR #292, opened 2021-03-19
fix poller launchctl wrapper call

escapewindow:
This moves the path definition earlier, so we can populate the template with the path.
Tested on mac-v3-signing14.


bhearsum (approved):

escapewindow:
Thanks!

PR #293, opened 2021-03-19
1804: network time tweaks

aerickson:
18.04 (and 16.04) by default don't use ntp for network time, but timesyncd.

- Place a timesyncd config file with our local NTP server.


aerickson:
timesyncd seems to detect that ntpd is running and disabled itself, there's an explicit command to do so that we probably should have done, but thing seem working.

`timedatectl set-ntp no`

This PR seems unnecessary now.

PR #294, opened 2021-03-23
Add Windows Moonshot tools and Windows logging reduction

markcor:
See win_moonshot_tools.md for tools explanation. 

Adjust task user init file to reduce logging.

Set Windows Moonshot worker timeout to 12 hours to reduce logging and hangs.


dividehex (changes requested):
Simply fix the bash script to allow tests to pass.   Otherwise, all looks good.

dividehex (approved):

PR #326, opened 2021-06-22
update r10k modules

aerickson:


aerickson:
@markcor Since we don't have testing on Windows, thoughts on landing this and reverting if there are issues?

Macs are using Bolt for deployment now (no scheduled convergence, just on demand) so breakage won't be immediate on landing.

dividehex (dismissed):
Let's make sure we get sign off from @markcor

PR #327, opened 2021-06-23
Signing data

dividehex:
This PR introduces hiera role data for the 3 signing/notarization roles.  In addition this fixes the following:

- inverts the OS version logic on hg symlinking
- logging defaults to syslog mdc1 service


aerickson (approved):
lgtm

davehouse (approved):

PR #330, opened 2021-07-12
Add vault agent to mac signing hosts

dividehex:
Deploystudio already adds the vault agent but this module will manage it and update the agent.


bhearsum (dismissed):
rubberstamp+

davehouse (approved):

PR #333, opened 2021-07-13
Fix papertrail lookup

dividehex:
The problem here is looking up an Integer in vault works fine when we look it up directly (eg. `vault_secrets::papertrail.data.port`) but we are actually looking up `papertrail.port` which hiera is then doing an interpolation lookup to vault.  In that case, interpolation converts and returns a String.  So we need to cast the string to an integer.  The integer function should also accept an integer and return an integer if that happens to be the case.  That would happen if the vault lookup returned empty and hiera would then returns what is defined in the vault.yaml.  If neither vault, nor vault.yaml contain a value, the lookup function returns the default `$syslog_port`

This is untested but I'm pretty confident it will fix the issue reported by @escapewindow


escapewindow (approved):
Thanks!

Issue #332, opened 2021-07-12
puppet broken on mac signers?


escapewindow:
> Error: Evaluation Error: Error while evaluating a Function Call, [hiera-vault] no token set in options and no token in VAULT_TOKEN (file: /etc/puppet/environments/production/code/modules/roles_profiles/manifests/profiles/logging.pp, line: 56, column: 41) on node mac-v3-signing9.srv.releng.mdc1.mozilla.com
curl: (22) The requested URL returned error: 401

1) How do we fix this? Do we have to reimage the entire pool?

2) I wish we could somehow keep track of things like this. Per [this comment](https://github.com/mozilla-platform-ops/ronin_puppet/pull/328#issuecomment-878504517), I manually puppetized one mac signer to see if things would work, but I could see us someday rolling out a change to `production-mac-signing` that bricks the entire mac signing pool.

escapewindow:
@dividehex

dividehex:
> 
> 
> > Error: Evaluation Error: Error while evaluating a Function Call, [hiera-vault] no token set in options and no token in VAULT_TOKEN (file: /etc/puppet/environments/production/code/modules/roles_profiles/manifests/profiles/logging.pp, line: 56, column: 41) on node mac-v3-signing9.srv.releng.mdc1.mozilla.com
> > curl: (22) The requested URL returned error: 401
> 
>     1. How do we fix this? Do we have to reimage the entire pool?

No need to reimage.  It should be fixed with PR 330.  The puppet script should pull the fix and run puppet again.

>     2. I wish we could somehow keep track of things like this. Per [this comment](https://github.com/mozilla-platform-ops/ronin_puppet/pull/328#issuecomment-878504517), I manually puppetized one mac signer to see if things would work, but I could see us someday rolling out a change to `production-mac-signing` that bricks the entire mac signing pool.

Maybe we should talk about this over zoom.  "Bricking" is always a risk and there are ways to mitigate it but we will need to collaborate and plan something out.

dividehex:
#330

escapewindow:
I applied the run-puppet.sh changes on mac-v3-signing9, and now I'm getting

```
[...]

PR #335, opened 2021-07-15
roll out latest scriptworker + scriptworker-scripts

escapewindow:
We're still seeing #332 bustage using vault, so I'm planning on cherry-picking this revision into the production-mac-signing branch to unblock esr91 without bricking the pool.


escapewindow:
@jcristau

bhearsum (approved):

PR #337, opened 2021-07-16
Do not change approle file content

dividehex:
🤦


dividehex:
TLDR;  vault_agent module would over write the approle credential files causing the vault agent daemon to fail to renew the token sink.

escapewindow (approved):

PR #339, opened 2021-07-17
MacOS Tests

dividehex:
This PR is to add integration testing for the handful of macos roles we support.  I think this branches current state is a good place to merge at even though it still lacks through test coverage in the form of tests.  This puts most of the infrastructure in place to run those test.  Another big issue I need to over come in a near future RP is the amount of repetition test code.  I'll need to come up with a solution to reuse common tests among OS/Role combos.  There is a lot of room for DRY improvements.

But in the mean time, let's get this reviewed and merged to prevent bit rot.

ps. I intend to squash all the commits in this PR before merging.


data/roles/gecko_t_osx_1100_m1.yaml
```diff
@@ -30,7 +30,7 @@ packages_classes:
   - python3
   - python3_psutil
   - python3_zstandard
-  - rosetta_2
+#  - rosetta_2
```
davehouse, line 33: I think we still need to install rosetta on the m1's. (still x86 are: python installers (installs universal runtime), puppet-agent)

davehouse (approved):
:fireworks: this is wonderful to get real tests for mac puppet!

data/roles/gecko_t_osx_1100_m1.yaml
dividehex: This is true but forcing the install fails on non-arm hardware which is what the circleci vm instances run.  So we don't really have proper arch coverage on that platform for now.  I think it is acceptable to not include it here by the simple fact that installing puppet itself from the pkg requires rosetta2 to be installed.  So part of the bootstrapping process of an M1 is to install rosetta2 and then install puppet.

aerickson (approved):
Looks amazing. Nice! LGTM.

data/roles/gecko_t_osx_1100_m1.yaml
davehouse: :+1: good reasoning. we'll have installed it when bootstrapping, or have to reinstall it to apply puppet again if it gets removed

PR #344, opened 2021-08-04
Include VAULT env in puppet apply script (to be merged direcly to production-mac-signing)

dividehex:
This PR is a single file fix to the run-puppet.sh script on darwin.  It matches what is on the master branch already.  The plan is to merge this directly to the production-mac-signing branch and then manually puppet apply each signing host against that branch in order to fix the current broken puppet state.  Once all the signing hosts are running puppet successfully, we should be able to safely merge master into the production-mac-signing branch.


escapewindow:
Thank you @dividehex !

I'll quarantine a mac, and:
- turn off auto-puppet
- apply this patch
- puppetize
- update to master
- puppetize

If those go well, then we should be good.

escapewindow (approved):
> Thank you @dividehex !
> 
> I'll quarantine a mac, and:
> 
>     * turn off auto-puppet
> 
>     * apply this patch
> 
>     * puppetize
> 
>     * update to master
> 
>     * puppetize
> 
> 
> If those go well, then we should be good.

Worked! I updated back to production-mac-signing and reenabled auto-puppet on the box as well. I think we should be good with this fix, thank you!

PR #345, opened 2021-08-05
roll out scriptworker 39.0.0

escapewindow:


escapewindow:
(We might not roll this out til I'm back from PTO, but adding more verbose info just in case, + posterity.)

When we roll out, this PR will stop marking the old CoT keys as valid. In theory I got everything working, but it's this may affect repos+branches where I missed rebuilding cached tasks, or older release graphs where we want to, say, do off-cycle partner repacks where we'll depend on older tasks.

If we hit CoT errors around signatures, try to trace back to the offending taskId (should be in the log). This will likely be a cached task. Rebuilding all the cached tasks on that repo/branch/level should fix the problem, though we'll need a rebuild of the relevant shippable build(s) afterwards. To do so:

```
# find the decision task
# download full-task-graph.json
# get the labels of the cached tasks:
gron full-task-graph.json | grep cached_task | cut -d . -f 1 | sort | uniq | sed -e 's/json\["//' -e 's/"\]//' | tee cached_tasks.txt
# go back to that decision task's task group, `add-new-jobs` those labels
```

Alternately, we can back out the scriptworker 39.0.0 rollout, or alternately craft a new scriptworker release after backing out [this commit](https://github.com/mozilla-releng/scriptworker/commit/1b13b1b0628b269b7fd75f12514671bfb3143cfc), and roll that out.

Rolling out to the `production-mac-signing` branch will roll out to all mac signers.

escapewindow:
Closing until we can properly investigate scriptworker + tc.py>39 issues.

PR #352, opened 2021-09-10
macmini-r8-{179..182} for mozvpn

davehouse:
179,180 in it42
181,182 in it44


inventory.d/signing.yaml
```diff
@@ -30,3 +30,15 @@ groups:
       - dep-mac-v3-signing3.srv.releng.mdc1.mozilla.com
     facts:
       puppet_role: mac_v3_signing_dep
+  - name: mac-v3-signing-mozillavpn
+    targets:
+      - macmini-r8-180.test.releng.mdc1.mozilla.com
+      - macmini-r8-182.test.releng.mdc1.mozilla.com
+    facts:
+      puppet_role: mac_v3_signing_mozvpn
+  - name: mac-v3-signing-mozillavpn
```
bhearsum, line 39: I think these two are just for building? If so we should probably drop `signing` from this name + the role (it took me an awful long time to figure out what why we had two `signing` entries...)

davehouse: sorry for the confusion. I was using the name for the puppet role and expecting they were using the "signing" scriptworker instead of another taskcluster worker.

aerickson:
Old. Closing.

PR #354, opened 2021-09-22
bug1731655 gsvelto m1 loaner

davehouse:


davehouse:
loaner was returned (and I'm reclaiming it)

PR #355, opened 2021-09-29
Bug 1733144: install python*-testresources apt packages

aerickson:
https://bugzilla.mozilla.org/show_bug.cgi?id=1733144


aerickson:
Tested in production on 001. Converged fine a few times.

PR #357, opened 2021-10-13
add hneiva to releng signing user list

escapewindow:
Heitor will be another Mac Signing maintainer on the Releng side.


hneiva (dismissed):
LGTM

PR #369, opened 2022-01-05
Fix Python dependency issues in mac signing workers

ahal:
I was seeing dependency issues when running puppet on the mac signers. Pinning jsmin >= 3 and updating dependencies seems to resolve it.


escapewindow (approved):
Thank you!

PR #368, opened 2021-12-28
Remove bhearsum and mtabara from mac signers

escapewindow:
We may add more releng in the future.

PR #371, opened 2022-01-24
Add mac_notarize_vpn behavior + update releng team

hneiva:


escapewindow:
As mentioned in channel:
- this PR updtes the gecko mac signing pool, but we probably want to create a vpn signing pool; and
- we probably want a larger conversation about who should have access to the mac signers. Added to the IT meeting notes.

hneiva:
- Users will be updated on #368
- will use a separate pool for vpn

PR #384, opened 2022-05-11
add releng users to test+build workers

davehouse:


modules/roles_profiles/manifests/profiles/signing_users.pp
```diff
@@ -10,11 +10,12 @@
             # That is where the user virtual resources are generated
             require roles_profiles::profiles::users
             # Lookup the group array and realize their user resource
-            $users = lookup('user_groups.signing', Array, undef, undef)
+            $users = lookup('user_groups.signing', Array, undef, undef) - lookup('user_groups.relops', Array, undef, undef)
```
escapewindow, line 13: Am I misreading, or will this result in 0 signing users?

markcor (approved):

modules/roles_profiles/manifests/profiles/signing_users.pp
escapewindow: Oh, hang on, we're already creating the `relops` users so we'd be adding the `signing` users twice?

escapewindow (approved):

modules/roles_profiles/manifests/profiles/signing_users.pp
gcoxmoz: [derail] While virtual resources are icky, in theory you should be able to realize them multiple times.  I would love to see stacktrace this causes, if you can simulate a failure, and see if we can eliminate it.
(In infra-puppet we gave up on virtual resources altogether, but I've stared at this class of problems a lot over time).

ahal (approved):

hneiva (approved):

ahal:
@aerickson @davehouse Are there any other blockers here? If we had access we could help alleviate some load by looking into problems like:
https://firefox-ci-tc.services.mozilla.com/tasks/a-HvPS6yR_OubO7x_H6aIA/runs/0/logs/public/logs/live.log

aerickson (approved):
lgtm

I'm fine with either approach. We could remove this change in a followup PR that add Releng directly like you mention.

PR #387, opened 2022-05-19
VPN Mac Puppetization

hneiva:
Set up puppet for VPN


modules/roles_profiles/manifests/profiles/mac_v3_signing.pp
```diff
@@ -30,6 +30,9 @@
             $role = $facts['networking']['hostname']? {
                 /^dep-mac-v3-signing\d+/ => 'dep',
                 /^tb-mac-v3-signing\d+/ => 'tb-prod',
+                # TODO: Remove 18 (dev machine)
+                # Hard-coded VPN hosts
+                /^mac-v3-signing(18|19|20)$/ => 'vpn',
```
hneiva, line 35: - [ ] Ask relops to change these workers' hostnames
- [ ] Update regex to match new hostnames

data/common.yaml
```diff
@@ -37,7 +37,7 @@ scriptworker_users:
       worker_id_suffix: "a"
       cot_product: "firefox"
       widevine_filename: "widevine_dep.crt"
-      keychain_file: "certs/dep-signing.keychain"
+      keychain_filename: "dep-signing.keychain"
```
escapewindow, line 40: 👍

data/common.yaml
```diff
@@ -117,6 +117,28 @@ scriptworker_users:
...
+      home: /builds/scriptworker
+      supported_behaviors:
+        - mac_notarize_vpn
+      notarization_users:  # Do we need this?
+        - notary1
+      dmg_prefix: "prod"
+      cot_product: "mozillavpn"
+      keychain_filename: "mozilla.20220330.keychain"
+      widevine_filename: "N/A"
+      ed_key_filename: "N/A"
```
escapewindow, line 131: If we want to e.g. beetmove a signed binary, we'll need to CoT-sign our CoT artifact, so we'll need this specified for prod mac signers.

escapewindow (commented):
This is looking good -- how is it working?

modules/signing_worker/manifests/init.pp
```diff
@@ -210,37 +213,41 @@
...
-    }
-    exec { "install ${scriptworker_base} widevine":
-        command     => "${virtualenv_dir}/bin/python setup.py install",
-        cwd         => $widevine_clone_dir,
-        user        => $user,
-        group       => $group,
-        refreshonly => true,
-        subscribe   => [Exec["clone widevine ${scriptworker_base}"], Python::Virtualenv["signingworker_${user}"]],
-        require     => [Exec["clone widevine ${scriptworker_base}"], Python::Virtualenv["signingworker_${user}"]],
+    if $widevine_filename {
```
hneiva, line 216: I changed it so we check if `$widevine_filename` is set instead of only doing it for ff role.
Once we confirm [or not] that TB doesn't need it, we can just delete the config.

ahal (dismissed):
Lgtm, but this is a rubberstamp. Consider blocking on Aki's review if unsure.

hneiva:
Vault changes:

- [ ] signing_secrets.vpn.release.{notarization_account,notarization_password,identity,keychain_password,pkg_cert_id}

data/common.yaml
```diff
@@ -117,6 +119,26 @@ scriptworker_users:
...
+      supported_behaviors:
+        - mac_notarize_vpn
+      notarization_users:  # Do we need this?
+        - notary1
+      dmg_prefix: "prod"
+      cot_product: "mozillavpn"
+      keychain_filename: "mozilla.20220330.keychain"
+      ed_key_filename: "ed25519_privkey"
+    # TODO: Split dep into it's own worker
+    depbld1:
```
hneiva, line 134: Unfortunately having dep on the same worker will require extensive changes to puppet logic. As is, the current setup doesn't support having 2 users (dep/prod) with different secrets.
ie: cltbld{1,2,3} will all have the same secrets, cannot have cltbld and dep under the same role.

@escapewindow we could move vpn dep to the dep workers along with ff and tb. Thoughts?

escapewindow: Can do, I think. It might be nice to have memory/cpu nagios alerts set up so we know if the dep boxes are getting overloaded.

data/common.yaml
```diff
@@ -61,8 +61,18 @@ scriptworker_users:
       worker_type_prefix: "tb-"
       worker_id_suffix: "tb"
       cot_product: "thunderbird"
+      # TODO: confirm if we need widevine for tb
       widevine_filename: "widevine_dep.crt"
       keychain_filename: "dep-signing.keychain"
+    vpnbld:
```
hneiva, line 67: moved dep building to dep worker now.
Haven't tested yet, will have to quarantine a dep worker to test it out.

escapewindow: We're much more resource-constrained on the dep pools, so it may be safer to puppetize a prod worker while forcing it to use dep configs.

data/common.yaml
```diff
@@ -61,8 +61,18 @@ scriptworker_users:
...
+      home: /builds/vpn-dep
+      supported_behaviors:
+        - mac_sign_and_pkg_vpn
+      dmg_prefix: "dep1"
```
escapewindow, line 71: I believe we have these `dmg_prefix`es to avoid multiple scriptworkers on the same host from trying to clobber each other. https://github.com/mozilla-releng/scriptworker-scripts/blob/3fc292ffca72364787d2ddd7e7c6e1220b2d7f54/iscript/src/iscript/mac.py#L520

(Maybe we should use a tempdir?)

That would suggest we want the dmg_prefix to be `vpn` or `vpn1` instead of conflicting with depbld1's `dmg_prefix`.

data/common.yaml
```diff
@@ -76,6 +86,7 @@ scriptworker_users:
         - mac_notarize_part_3
       dmg_prefix: "prod"
       cot_product: "thunderbird"
+      # TODO: confirm if we need widevine for tb
```
escapewindow, line 89: might be worth asking @jfx2006.

data/common.yaml
```diff
@@ -117,6 +128,17 @@ scriptworker_users:
         - notary8
         - notary9
         - notary10
+  vpn:
+    cltbld:
+      home: /builds/scriptworker
+      supported_behaviors:
+        - mac_notarize_vpn
+      notarization_users:  # Do we need this?
+        - notary1
```
escapewindow, line 137: Really only needed if we want to start using https://github.com/mozilla-releng/scriptworker-scripts/blob/3fc292ffca72364787d2ddd7e7c6e1220b2d7f54/iscript/src/iscript/mac.py#L719 again.

We tend to get rate limited that way, so we found submitting one big zip is more reliable. We haven't needed sudo-wrapped notarization in a while, so we could potentially remove it from iscript.

escapewindow (approved):
Looks good except for the `dmg_prefix`, ty!

data/common.yaml
```diff
@@ -61,8 +61,18 @@ scriptworker_users:
       worker_type_prefix: "tb-"
       worker_id_suffix: "tb"
       cot_product: "thunderbird"
+      # TODO: confirm if we need widevine for tb
```
escapewindow, line 64: IIRC we can drop this comment?

escapewindow (approved):

data/common.yaml
hneiva: Removed and rebased

ahal (approved):

PR #388, opened 2022-05-20
Fix depbld1 keychain filename variable

hneiva:


hneiva:
I have this change on my other PR, but let's add this sooner than later

ahal (dismissed):
Lgtm, but wouldn't mind an explanation around why this fixes the deploy issues we were seeing. Consider adding said explanation to the commit message :)

hneiva:
The variable name is `keychain_filename`, not `keychain_file`. Also, the `certs/` part of the path is wrong.
I believe this was a slip on my part when we rolled new certs

ahal (approved):

PR #389, opened 2022-05-25
Minor fix for SMTPException and better logging

hneiva:


escapewindow (dismissed):
Works for me; hoping this long-standing error gets fixed, ty!

Note: puppet-darwin-run-puppet.sh is used by *all* macs managed by ronin-puppet, including mac hardware builders and mac hardware testers. This is worth flagging @davehouse for review.

modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
```diff
@@ -94,7 +94,7 @@ try:
    smtpObj = smtplib.SMTP('<%= @smtp_relay_host %>')
    smtpObj.sendmail("${SENDER}", "${RECEIVER}", msg)
    print "Successfully sent email"
-except SMTPException:
+except smtplib.SMTPException:
    print "Error: unable to send email"
```
gcoxmoz, line 98: Why catch the exception at all?   You're at the tail end of a short embedded python script that's about to exit (whether it succeeds or fails) and throw control back to a bash script that doesn't seem to be `-e`.  The `except` hides the traceback and is making sending issues undiagnosable.  All this seems to get you is brevity on failures.

hneiva: That's a good point.
If no one objects, I'll remove the try/except.

modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
```diff
@@ -78,8 +78,15 @@ function email_report {
...
     ERR_MSG=$2
 
+    RELAY_HOST="<%= @smtp_relay_host %>"
     SENDER="root@${FQDN}"
     RECEIVER="${PUPPET_MAIL}"
+
+    if [[ "${RELAY_HOST}" == 'localhost' ]]; then
+        echo "No email server setup. Skipping email."
+        echo $ERR_MSG
+        return
```
hneiva, line 88: @gcoxmoz @escapewindow I was thinking about this, I think it's safe to assume we don't have localhost smtp on the macs.
So I exit early if it's localhost, and don't catch the exception in the python otherwise.

gcoxmoz: I can't review because I can't confirm/deny your assumption, but I like the thought process if you're right.

escapewindow: It's certainly possible to set up localhost be a relay to a real mail server. Not sure if the hardware mac builders have localhost smtp set up or not, or if we want them to. It would certainly make alerting easier if we had a .forward that sent mail to release+something@m.c.

hneiva:
Fixed how the SMTP host value is loaded now. Thanks to aerickson for helping me debug.

escapewindow (approved):

PR #394, opened 2022-06-29
Update iscript requirements + refactor update script

hneiva:
Please note I replaced the use of `sed` to interpret the yaml file with `yq`
The `sed` search broke with the addition of VPN, and `yq` properly parses the data instead of raw text searching it


escapewindow (approved):
I don’t love using the yq… we’ve probably outgrown bash and should look at python. Can be a followup.

PR #401, opened 2022-07-12
add jcristau and jlorenzo

escapewindow:


escapewindow:
@davehouse could we also add them to relops duo? (should I file a ticket?)

escapewindow:
It also looks like I can't invite them to the github org; I may only be able to add existing org users to the releng team.

escapewindow:
https://mozilla-hub.atlassian.net/browse/RELOPS-215

PR #402, opened 2022-07-15
Bug 1774669 - add mac_notarize_single_file behavior

gbrownmozilla:


escapewindow (approved):
This will take effect when we push to the production-mac-signing branch.
If we want to update mac-v3-signing20 for adhoc, we'll need to do so manually (ssh in and edit /builds/scriptworker/script_config.yml) since adhoc isn't supported in puppet currently.

escapewindow (approved):

PR #403, opened 2022-07-18
Don't loop forever in run-puppet.sh on mac

jcristau:
When using the puppet::periodic class we can end up in a situation where
run-puppet.sh is buggy and instead of fixing itself on a later periodic
run when it pulls a fix from git, it just keeps failing until it's
manually killed or the host gets rebooted.

So limit the loop to 5 iterations, then give up.

### DIFF
--- a/modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
+++ b/modules/puppet/templates/puppet-darwin-run-puppet.sh.erb
@@ -183,10 +183,18 @@ function run_puppet {
 
 block_on_network
 
-# Call the run_puppet function in a endless loop
-while ! run_puppet; do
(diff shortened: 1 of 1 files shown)


escapewindow (approved):
Haha, /me sees `/var/tmp/semaphore/run-buildbot` below.

I think this is worth a shot. If we figure out that having it be an endless loop is good for first-time post-reimage, if puppet is busted and we later land a fix (after 5 retries have failed), we could look at the existence of `/var/tmp/semaphore/run-buildbot` or similar as a way to determine whether we've successfully puppetized before. If so, limit the max attempts; if not, keep looping.

(Then we'd include a "check to see if puppet is still running" in our manual-post-reimage steps docs.)

PR #406, opened 2022-07-28
Ruby 3 compatibility

aerickson:
Jonathan ran into issues trying to get test-kitchen setup on Ubuntu 22 via WSL. This should make it work.

- update gems to work with Ruby 3
- change CCi config to use Ruby 3 for Windows


jwmoss:
Sweet! When I get back from PTO I’ll give this a shot.

PR #411, opened 2022-08-15
Win 2012 initial support 

markcor:


markcor:
@jwmoss : 

Majority of the files are from updating r10k modules. 

Pulled updates from master branch. That is the cause of the diff with modules/puppet/templates/puppet-darwin-run-puppet.sh.erb and test/integration/linux/inspec/users_spec.rb.  

provisioners/windows/azure/azure-bootstrap.ps1, seems like the most of the diff is a matter of linting. let's get it into the cloud_windows branch and then re-lint.

jwmoss (changes requested):
in `azure-maintainsystem.ps1` line 359, if a Z: drive doesn't exist then skip it. Something like:

```Powershell
if ((Get-Volume).DriveLetter) -contains Z) {
  Write-Log -message ('{0} :: Opening permisions for Z:.' -f $($MyInvocation.MyCommand.Name)) -severity 'INFO'
  icacls.exe "Z:\" /grant "Everyone:(OI)(CI)F"
}
```

markcor:
> in azure-maintainsystem.ps1 line 359, if a Z: drive doesn't exist then skip it. Something like:

That line actually has no effect. Removing it. It looks like the `Set-DriveLetters` could maybe use some future clean up.

jwmoss (approved):

PR #414, opened 2022-09-01
test fixing

aerickson:
Run OS X 10.14 builds on 10.15 since no 10.14 images are available on CCI. See https://mozilla-hub.atlassian.net/browse/RELOPS-235 and https://circleci.com/docs/testing-ios#supported-xcode-versions for more info.

Also short-circuits/auto-passes the windows tests on forked PRs (they will always fail due to no secrets being available). 
- The solution to get testing on Windows changes is to push the branch to mozilla-platform-ops/ronin-puppet and create a PR from the repo (then the tests can access secrets).
- We could also switch to windows docker images on windows executors (which wouldn't require secrets/credentials). https://circleci.com/docs/using-windows#windows-machine-executor-images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,18 @@ executors:
       xcode: 12.5.1
 
 commands:
+  early_return_for_forked_pull_requests:
+    description: >-
+      If this build is from a fork, stop executing the current job and return success.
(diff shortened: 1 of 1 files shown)


aerickson:
Discussed with jwmoss.

PR #416, opened 2022-09-19
Add support for azure utilities

jwmoss:
Adding support for azure utilities. Initial upload includes scripts to check for license status.


markcor:
In case like this where we have a base script and supporting scripts, let's put them in their own directory, so in this case something like `utilities/windows/azure/kmsaudit` .

markcor (approved):

PR #423, opened 2022-10-11
Initial support for disabling UAC

jwmoss:


modules/roles_profiles/manifests/profiles/disable_uac.pp
```diff
@@ -0,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class roles_profiles::profiles::disable_uac {
```
markcor, line 5: Let's move this to `roles_profiles::profiles::suppress_dialog_boxes`  and put it under a Win 11 specific case statement since it has not been tested on the other worker pools.

markcor (changes requested):

modules/win_disable_services/manifests/disable_uac.pp
```diff
@@ -0,0 +1,17 @@
...
+class win_disable_services::disable_uac {
+  case $facts['custom_win_os_version'] {
+    'win_11_2009':{
+      registry_value { 'HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System\EnableLUA':
+        type => dword,
+        data => '0',
+      }
+    }
+    default: {
+      fail("${module_name} does not support ${$facts['custom_win_os_version']}")
```
markcor, line 14: Let's change this from `fail` to `warning`. That way it won't break images builds for existing pools. Though in the long term we should add this to all pools.

markcor (changes requested):

markcor (approved):

PR #436, opened 2022-10-19
Use docker to pin mac pip requirements

hneiva:
Prevents python version mismatching

### DIFF
--- a/modules/signing_worker/files/update-requirements.sh
+++ b/modules/signing_worker/files/update-requirements.sh
@@ -5,8 +5,8 @@
 # It will also fetch yq and python images which can be safely removed from the local image cache
 
 # Configuration
-yq_docker_image=mikefarah/yq:4.28.2
-python_docker_image=python:3.8.3-alpine3.12
+yq_docker_image=mikefarah/yq:4.28.2@sha256:473c4bc63c7b36793b4dbc980e183026d5422cb1ec6fe1f1cc12f1bc1b2e8be9
+python_docker_image=python:3.8.3-alpine3.12@sha256:6c1b18373c4f94353308097772e97ff4d0e596ac3d06cc4b00c3bdc52cd5e8b6
 
 # Note that MY_DIR is the location of the script...
 pushd "$(dirname "$0")" &>/dev/null || exit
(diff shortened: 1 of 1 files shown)


escapewindow:
I'd prefer the no-args use case to pin all `.in` files in the dir as a UX improvement. Not a blocker.

modules/signing_worker/files/update-requirements.sh
```diff
@@ -1,11 +1,12 @@
...
-    echo "Worker type is ff-prod, tb-prod, or dep"
-    echo "Pinned requirements will be output to stdout"
-    exit 1
-fi
+# This script will create a docker image on the local machine with name "pip-compile"
+# It is safe to delete such image after running this script
+# It will also fetch yq and python images which can be safely removed from the local image cache
+
+# Configuration
+yq_docker_image=mikefarah/yq:4.28.2
```
jcristau, line 8: I sort of understand using docker to force the python version, but yq should presumably be just fine to keep running outside of docker?

modules/signing_worker/files/update-requirements.sh
```diff
@@ -1,11 +1,12 @@
...
+python_docker_image=python:3.8.3-alpine3.12
```
JohanLorenzo, line 9: Big 👍 for making this script reproducible across all types of setups! If by any chance you would like to go one step further, we can even pin the docker image by hash:

```suggestion
python_docker_image=python:3.8.3-alpine3.12@sha256:6c1b18373c4f94353308097772e97ff4d0e596ac3d06cc4b00c3bdc52cd5e8b6
```

https://hub.docker.com/layers/library/python/3.8.3-alpine3.12/images/sha256-6c1b18373c4f94353308097772e97ff4d0e596ac3d06cc4b00c3bdc52cd5e8b6?context=explore

modules/signing_worker/files/update-requirements.sh
```diff
@@ -15,17 +16,39 @@ popd &>/dev/null || exit
 # be different
 cwd=$(pwd)
 
+if [ -z "$1" ]; then
+    echo "Updating all requirement files"
+    for f in ${MY_DIR}/requirements.*; do
+        target=$(echo $f | awk -F '.' '{print $2}')
+        echo "Updating ${f}"
+        ${MY_DIR}/${0} $target $f
```
JohanLorenzo, line 24: Nit: Let's quote args in order to avoid shell injection.

```suggestion
        "${MY_DIR}/${0}" "$target" "$f"
```

modules/signing_worker/files/update-requirements.sh
```diff
@@ -15,17 +16,39 @@ popd &>/dev/null || exit
...
+    done
+    exit
+fi
+
+output_file="requirements.${1}.txt"
+
 trap 'cd $cwd' EXIT
 
 worker_type=$1
```
JohanLorenzo, line 33: Oh, that answers my question about `$1` 🙂 Then let's move this line before we test the content of `$1`

JohanLorenzo (approved):
Thank you very much for making this script more consistent across setups ! 👍 

The changes look great to me module a few nits 🙂

modules/signing_worker/files/update-requirements.sh
ahal: So I had an older version of `yq` on my system and it failed to parse the files ><, upgrading to 4.28.2 fixed it (unsure what the minimum version is tho).

ahal (dismissed):

hneiva:
Refactored the file to target all files in the directory primarily (which should be our ~90% primary use case anyways)

Thank you guys for the suggestions!

modules/signing_worker/files/update-requirements.sh
hneiva: > I sort of understand using docker to force the python version, but yq should presumably be just fine to keep running outside of docker?

As ahal mentioned, there are minor differences with the versions, and in some systems `yq` installation (mac vs debian vs arch etc) can be a pain for the developer.
The `yq` docker image here is 18MB and very fast

modules/signing_worker/files/update-requirements.sh
```diff
@@ -15,34 +16,65 @@ popd &>/dev/null || exit
...
+if [ -z "$cli_param" ]; then
+    worker_types=()
+    echo "Updating all requirement files"
+    for f in ${MY_DIR}/requirements.*; do
+        target=$(echo $f | awk -F '.' '{print $2}')
+        echo "Found target: ${target}"
+        worker_types+=("${target}")
+    done
+else
+    worker_types=($cli_param)
```
JohanLorenzo, line 30: TIL the `()` operator!

modules/signing_worker/files/update-requirements.sh
```diff
@@ -53,5 +85,23 @@ cryptography
...
+        pip-compile:latest \
+        pip-compile -q --generate-hashes -o $output_file -r requirements.in
+    
+    cp "${workdir}/${output_file}" "${MY_DIR}/${output_file}"
+    rm "${output_file}" "requirements.in"
+}
+
+for wt in "${worker_types[@]}"; do
+    echo "Compiling $wt"
+    compile $wt
```
JohanLorenzo, line 104: Nit: 

```suggestion
    compile "$wt"
```

JohanLorenzo (approved):
LGTM!

PR #438, opened 2022-10-26
Update signing_worker dependencies

ahal:


modules/signing_worker/files/requirements.dep.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --allow-unsafe --generate-hashes requirements.in
+#    pip-compile --generate-hashes --output-file=requirements.dep.txt requirements.in
```
escapewindow, line 5: We're no longer using `--allow-unsafe`; is this a desired change?

escapewindow (approved):

modules/signing_worker/files/requirements.dep.txt
ahal: From matrix:
> yes, we aren't pinning any packages that are considered unsafe anyway so it's a no-op (and a result of me using Heitor's new script)

PR #447, opened 2022-11-02
linux: tc component update and py39

aerickson:
main changes:
- update TC components to v44.23.2
- use alternatives to set /usr/bin/python3 to py3.9 (vs 3.6 currently/default)
- upgrade to newer version of 3.9
  - we didn't store a package that's now needed for the current/older version and it's not available now

other changes:
- stop installing python3.8
  - not needed with 3.9, and was messing with 3.9 installation
- update py3 versions of zstd and psutil
  - required to work with py3.9
- upgrade pip3 to latest 
- update versions of puppet and bolt installed in mac CircleCI tests
  - not needed, but did as part of debugging mac test failures and seems harmless

see https://mozilla-hub.atlassian.net/browse/RELOPS-239 and https://mozilla-hub.atlassian.net/browse/RELOPS-246.


aerickson:
Testing done:
- test-kitchen: converges and inspec tests pass
- converged prod node 001, has been running for a day with no issues

PR #459, opened 2022-12-16
Update signing_worker dependencies

ahal:
This picks up scriptworker 46.0.0 which includes the 2021 CoT public key removal.

PR #465, opened 2023-01-18
Update mac signer scriptworker[,-scripts] deps

hneiva:
We should close #459 once this is merged.


hneiva:
Tested on `mac-v3-signing1`

ahal (approved):

PR #473, opened 2023-01-20
DO NOT MERGE - Create Ubuntu 22.04 for moonshots

MasterWayZ:


modules/roles_profiles/manifests/profiles/gecko_t_linux_talos_generic_worker.pp
```diff
@@ -20,11 +20,9 @@
             }
 
             # TODO: move these lines to linux-base?
-            require linux_packages::py2
```
aerickson, line 23: These still need to be installed on the 1804 configuration.

aerickson, line 23: These still need to be installed on the 1804 configuration.

MasterWayZ: Fixed, 2204 config has its own file now.

MasterWayZ:
Squashed all commits together so the first commit doesn't say 'do not land'.
This is ready for review now.

.circleci/config.yml
```diff
@@ -303,7 +303,7 @@ workflows:
       - linux_integration_tests:
           matrix:
             parameters:
-              kitchen_target: ["bitbar-ubuntu-1804", "bitbar-ubuntu-2204", "linux-ubuntu-1804"]
+              kitchen_target: ["bitbar-ubuntu-2204", "linux-ubuntu-1804"]
```
aerickson, line 306: we should have linux-ubuntu-2204 in here

MasterWayZ: Good catch!

data/roles/gecko_t_osx_1015_r8_loaner.yaml
```diff
@@ -8,7 +8,7 @@ worker_metadata:
   provisionerId: "%{lookup('gw.provisioner_id')}"
 
 worker:
-  taskcluster_version: 40.0.3
+  taskcluster_version: 47.0.2
```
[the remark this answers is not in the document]
MasterWayZ: Oh whoops! That didn't mean to make it in!

test/integration/linux/inspec/ntp_spec.rb
```diff
@@ -1,7 +1,15 @@
...
   its(:exit_status) { should eq 0 }
   # TODO: check clock sync... it doesn't work on freshly converged hosts though.
   its(:stdout) { should match /systemd-timesyncd.service active: yes/ }
 end
+
+only_if { os[:family] == 'ubuntu' && os[:release] == '22.04' }
+describe command('timedatectl status') do
+  its(:exit_status) { should eq 0 }
+
+  # TODO: check output if we're synced. For 22.04 the output has changed.
```
MasterWayZ, line 14: @aerickson I have no clue how we'd fix that. Done this so at least I can see it everything else passes, and it does!

inventory.d/macmini-r8.yaml
```diff
@@ -122,13 +122,6 @@ groups:
       - macmini-r8-117.test.releng.mdc1.mozilla.com
       - macmini-r8-118.test.releng.mdc1.mozilla.com
       - macmini-r8-119.test.releng.mdc1.mozilla.com
-      - macmini-r8-120.test.releng.mdc1.mozilla.com
```
aerickson, line 125: remove mac content

inventory.d/macmini-r8.yaml
```diff
@@ -265,13 +258,20 @@ groups:
 
   - name: gecko-t-osx-1015-r8-staging
     targets:
+      - macmini-r8-120.test.releng.mdc1.mozilla.com
```
aerickson, line 261: remove mac content

.kitchen_configs/kitchen.docker.yml
```diff
@@ -77,7 +84,13 @@ suites:
       custom_options: '-e "include roles_profiles::roles::gecko_t_linux_talos"'
     attributes:
     includes:
-      - ubuntu-18.04  # no moonshot 2204 config yet
+      - ubuntu-18.04
```
aerickson, line 87: I think we should have a single 'linux' suite block and override the custom options for provisioner in the platforms section like:

```
platforms:
  - name: ubuntu-18.04
  - name: ubuntu-22.04
    provisioner:
      custom_options: '-e "include roles_profiles::roles::gecko_t_linux_2204_talos"'
```

modules/roles_profiles/manifests/roles/gecko_t_linux_talos.pp
```diff
@@ -18,6 +18,15 @@
     #include ::fw::roles::linux_taskcluster_worker
 
     # talos stuff
-    include ::roles_profiles::profiles::gecko_t_linux_talos_generic_worker
-
+    case $::operatingsystemrelease {
```
aerickson, line 21: Do we need to change this / have an if block? Aren't we specifying the 2204 profile in the 2204 role profile?

test/integration/linux/inspec/services_disabled_spec.rb
```diff
@@ -30,8 +30,11 @@
   it { should_not be_enabled }
 end
 
-describe service('whoopsie') do
-  it { should_not be_enabled }
+# This check is broken on Ubuntu 22.04, even when the service is disabled.
```
aerickson, line 33: Is whoopsie even present/used in 2204?

inventory.d/macmini-r8.yaml
MasterWayZ: Oh dammit

.kitchen_configs/kitchen.docker.yml
MasterWayZ: Sure! Will edit!

modules/roles_profiles/manifests/roles/gecko_t_linux_talos.pp
MasterWayZ: They both point at this shared file. I might as well just make them its own files so 18.04 is easier to clean up

test/integration/linux/inspec/services_disabled_spec.rb
MasterWayZ: Yes, but from what I can see it is a static service so you cannot disable it.

modules/roles_profiles/manifests/roles/gecko_t_linux_talos.pp
MasterWayZ: This was put there because I thought something was broken. Turns out that the kitchen config was wrong so it never used the 2204 one. This change has been reverted.

test/integration/linux/inspec/services_disabled_spec.rb
MasterWayZ: @aerickson I manually ran `systemctl disable whoopsie` and it still fails on the test, so I think the test is broken or something changed that it works differently. On the raw `ubuntu:jammy` image whoopsie does not exist. I did see a `apt install` line for whoopsie in the output logs.. But I cannot find anything referencing it in the code.

aerickson: Setting it to disabled will make Puppet install it. Not intuitive, but it can't set disabled if the package isn't present.

Let's just do nothing regarding whoopsie on 2204 (ie in the 2204 block do nothing).

aerickson:
NTP solution is in my branch:
  - https://github.com/MasterWayZ/ronin_puppet/compare/relopsTODO...aerickson:ronin_puppet:fixes_to_relopsTODO?expand=1

I think you only need the changes to ntp.pp and ntp_spec.rb.

aerickson:
We also need to switch to multiuser g-w and create a configuration that uses x11 vs wayland (link in description on how to do).

PR #476, opened 2023-02-01
kitchen inspec: only enable color if not in CI

aerickson:
inspec output in CircleCI is colored red (https://github.com/inspec/inspec/issues/1700) which is very confusing.

Disable colored output if in CI. See https://github.com/inspec/kitchen-inspec/issues/103.


MasterWayZ (approved):
Thanks! This should make reading the output a lot easier!

PR #481, opened 2023-02-09
Remove dhouse

MasterWayZ:


test/integration/bitbar/inspec/bitbar_devicepool_spec.rb
```diff
@@ -8,13 +8,6 @@
     end
   end
 
-  describe user('dhouse') do
```
aerickson, line 11: I think we should replace with you vs delete. I guess slightly more involved than just changing the login here, so let's punt to another PR.

test/integration/linux/inspec/users_spec.rb
```diff
@@ -5,16 +5,11 @@
   describe user('aerickson') do
     it { should exist }
   end
-
```
aerickson, line 8: Replace with you or another team member vs delete.

test/integration/bitbar/inspec/bitbar_devicepool_spec.rb
MasterWayZ: Will make a new PR.

test/integration/linux/inspec/users_spec.rb
MasterWayZ: Done.

test/integration/bitbar/inspec/bitbar_devicepool_spec.rb
MasterWayZ: Made https://github.com/mozilla-platform-ops/ronin_puppet/pull/482

aerickson (approved):

PR #488, opened 2023-02-11
Make adjustments to a branch 1

MasterWayZ:
Note: `data/common.yaml` synced from master.
Most is just additions that are not really used. The scriptworker revision changing is something important though that we have to make sure is up-to-date.
For tests, it was easier to just sync the test files over. Tests passing or failing will tell if this works or not.

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -12,17 +12,17 @@ apple_firmware_acceptance:
 
 user_groups:
   relops:
-    - jwatkins
-    - dhouse
     - mcornmesser
     - aerickson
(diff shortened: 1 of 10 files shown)


MasterWayZ:
@aerickson bitbar tests are failling, this branch is not used by bitbar, just mac builders so we can ignore these.
The other failures, this branch has not been touched in forever so we are better off just making the edit and ignore CI for now.
In the future we should look into making all machines use the same branch (other than the signers).

aerickson (approved):

PR #491, opened 2023-02-14
RELOPS-425 - Update current running inventory

MasterWayZ:
Technically `modules/roles_profiles/manifests/profiles/gecko_3_t_osx_1014_generic_worker.pp` isn't part of the ticket but it's something we have to land anyway..
This comments out the known bad Macs and syncs the staging and prod definitions.


aerickson:
Add 246 in somewhere?

MasterWayZ:
@aerickson done. added to staging so I can check the machine before adding it to one of the other pools.

aerickson:
Why comment out the offline hosts? Just so we can get through a run on the entire fleet? 

I just worry we forget about the offline hosts and never bring them back, but works for now.

aerickson (approved):

MasterWayZ:
I am working on a sheet as I will be auditing all machines, so do not worry. I have a plan :)

PR #493, opened 2023-02-14
add bhearsum to mac signer access list

bhearsum:
I assume I need more than just this to get access, but this seems like step 1...


bhearsum:
(Will re-open this against the correct branch.)

PR #497, opened 2023-03-02
Migrate ronin puppet binaries from s3 to az

jwmoss:


data/os/Windows.yaml
```diff
@@ -27,19 +27,13 @@ windows:
   ext_pkg_src: "https://roninpuppetassets.blob.core.windows.net/binaries"
   ext_gpg_src: "https://roninpuppetassets.blob.core.windows.net/binaries/gpg"
 
-  # S3 resources
```
jwmoss, line 30: @markcor Should we remove this section yet?

markcor: Let's leave these in place until the moonshots are retired.

markcor: My bad. Because this is the cloud_windows branch, let;s go ahead and remove those.

markcor (approved):

PR #507, opened 2023-05-15
add safaridriver to r8 staging config

aerickson:
The staging r8 macs never had the safaridriver role/profile added to them. Fix that.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1812409 for more info.


MasterWayZ (approved):
Thank you!

PR #511, opened 2023-06-07
Bump iscript scriptworker-script revision + pin reqs

hneiva:


bhearsum (commented):
Is there a reason we're updating prod before deploying and testing with the dep machine? (AFAIK this isn't like the notarization changes, and can be fully tested with dep certs)

hneiva:
I've tested it in a quarantined mac manually. I already had it setup for testing other things, so it was the quickest way of getting it done.

bhearsum:
> I've tested it in a quarantined mac manually. I already had it setup for testing other things, so it was the quickest way of getting it done.

Ah, okay! As long as it's been tested already, sgtm

bhearsum (approved):

gbrownmozilla (approved):

PR #512, opened 2023-06-08
Sync local changes to fix puppet

MasterWayZ:
DONE:
- don't disable creation of cltbld
- reverted windows changes
- revert any windows changes
- fix pylint issues


MasterWayZ:
Windows changes reverted

data/roles/gecko_t_osx_1100_m1_staging.yaml
```diff
@@ -11,7 +11,7 @@ worker_metadata:
 worker:
   taskcluster_version: 44.0.0
   provider_type: standalone
-  worker_pool_id: releng-hardware/gecko-t-osx-1100-m1-staging
+  worker_pool_id: releng-hardware/gecko-t-osx-1100-m1
```
aerickson, line 14: We probably still want this in `-staging`.

inventory.d/hoststhatareskipped.yml
```diff
@@ -0,0 +1 @@
+218,219,220,221,224,226,228,229,231,235,236,240,241,243,248,250,252
```
aerickson, line 1: We probably don't want to check this in.

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -32,7 +32,7 @@ exec 2> >(tee /var/log/genericworker/stderr.log >&2)
 exitstatus="${PIPESTATUS[0]}"
 
 # Remove semaphore file to ensure generic worker does not start before puppet
-rm -f "/var/tmp/semaphore/run-buildbot"
+#rm -f "/var/tmp/semaphore/run-buildbot"
```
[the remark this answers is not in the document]
MasterWayZ: This is needed because for some reason we hit that rm intermittently, which causes workers to stop.
This was for puppet-on-boot which we don't have set up anyway. There are two pools running with this change and they work fine.

aerickson (approved):
lgtm

PR #532, opened 2023-07-27
Testing fix for bluetooth dialog

rcurranmoz:
Test fix for relops-321


MasterWayZ (approved):
This just affects a testing pool, but it is confirmed to set the right bits.
LGTM

PR #538, opened 2023-08-02
Bump iscript requirements + script to self-update

hneiva:


modules/signing_worker/files/update-requirements.sh
```diff
@@ -6,7 +6,9 @@
 
 # Configuration
 yq_docker_image=mikefarah/yq:4.28.2@sha256:473c4bc63c7b36793b4dbc980e183026d5422cb1ec6fe1f1cc12f1bc1b2e8be9
-python_docker_image=python:3.8.3-alpine3.12@sha256:6c1b18373c4f94353308097772e97ff4d0e596ac3d06cc4b00c3bdc52cd5e8b6
+# Note: Ultimately the Mac signers are going to install the resulting compiled requirements.
+#           If you're looking to update this image, make sure they can install it.
+python_docker_image=python:3.8.3-slim@sha256:938fd520a888e9dbac3de374b8ba495cc50fe96440030264a40f733052001895
```
hneiva, line 11: Quick note here, jsonschema wouldn't compile on alpine based images anymore, so I updated it to slim.
Reference: https://github.com/python-jsonschema/jsonschema/issues/1143

data/update-scriptworker-revisions.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
```
ahal, line 1: I feel like this should be under `modules/signing_worker` somewhere. Possible just alongside the requirements.txt files.

ahal: Ah sorry, I didn't realize that's where the revisions were. Why is that, do other modules need scriptworker?

ahal (approved):

data/update-scriptworker-revisions.py
hneiva: This is before my time, but for what I understand `/data/*` has all values to feed into the puppet scripts. So, pretty much every script will pull data from these files.

The update script only touches `data/common.yaml`, so I decided it should live right beside it. Same logic for the `modules/signing_worker/files/update-requirements.sh` script, as it's only modifying files in the same directory. Not sure if it would make sense for a script in `modules/signing_worker/` to modify `data/common.yaml`.

hneiva: Merging as is. If we decide to move this around, it should be in a separate PR.

PR #543, opened 2023-08-12
Bug1845790 reduce 2022 first VM bootup

markcor:


data/roles/win2022642009azurevs.yaml
```diff
@@ -0,0 +1,55 @@
+---
+
+win-worker:
+  # generic-worker
+  generic_worker:
+    # File versions
+    name: "generic-worker-multiuser-windows-amd64"
+    version: '53.2.1'
```
jwmoss, line 8: Did you want to bump this to `54.4.1`? That's what I'm rolling out with windows cloud workers (and hardware).

modules/roles_profiles/manifests/profiles/disable_services.pp
```diff
@@ -52,6 +52,10 @@
             if $facts['os']['release']['full'] == '10' {
                 include win_disable_services::disable_onedrive
             }
+            ## 2016 == 2022
+            if $facts['os']['release']['full'] == '2016' {
```
jwmoss, line 56: Does this translate to 2022 or 2016? Would it be better to have a custom fact instead?

jwmoss (changes requested):

data/roles/win2022642009azurevs.yaml
markcor: good call.

modules/roles_profiles/manifests/profiles/disable_services.pp
markcor: I am good with leaving it as is for now.  I want to get this out early next week to level 1/2 workers. However, we should refactor to `$facts['custom_win_os_version']` everywhere at some point.

jwmoss (approved):

PR #548, opened 2023-08-23
Update Safari bug fix

MasterWayZ:


modules/macos_safariupdate/manifests/init.pp
```diff
@@ -15,8 +15,6 @@
         command => $update_script,
         require => File[$update_script],
         user    => 'root',
-        # semaphore created in script
-        unless  => '/bin/test -f /Users/cltbld/Library/Preferences/semaphore/safari-update-has-run',
```
MasterWayZ, line 19: This only checks if the file exists, but it doesn't check if the right semaphore version is in there. There is a proper check in the shell script that does look at the version, so it should we safe to remove this.

rcurranmoz (approved):

PR #549, opened 2023-08-23
Add mozillavpn workers and update appservices

MasterWayZ:


modules/dirs/manifests/builds.pp
```diff
@@ -5,9 +5,6 @@
 class dirs::builds {
 
     $builds_dir = $facts['os']['macosx']['version']['major'] ? {
-        '13.5' => '/opt/builds',
```
MasterWayZ, line 8: These are redundant, as the default is set to `/opt/builds`

modules/dirs/manifests/tools.pp
```diff
@@ -5,9 +5,6 @@
 class dirs::tools {
 
     $tools_dir = $facts['os']['macosx']['version']['major'] ? {
-        '13.5' => '/opt/tools',
```
MasterWayZ, line 8: These are redundant, as the default is set to `/opt/tools`

modules/talos/manifests/init.pp
```diff
@@ -8,10 +8,8 @@
     case $facts['os']['name'] {
         'Darwin': {
             $builds_dir = $facts['os']['macosx']['version']['major'] ? {
-                '13.5' => '/opt/builds',
```
MasterWayZ, line 11: These are redundant, as the default is set to `/opt/builds`

MasterWayZ:
Note: moving appservices to ::worker is intentional

aerickson (approved):
LGTM. Very nice.

PR #558, opened 2023-09-22
fix semaphore dir race and add missing macs

aerickson:
The race condition wasn't noticed until now because my `safaridriver` class's enable script ran before the `safariupdate` class existed/ran on most macs, but these 'lost' macs didn't have the semaphore dir created as the correct user. This PR fixes that issue by chmodding in the earlier `safariupdate` class's script invocation.

RE: missing macs, Not sure how these were found, but rcurran sent me the list. They weren't converging (due to no role).


MasterWayZ (approved):
Looks good to me!

PR #561, opened 2023-10-02
Fix m1 config

MasterWayZ:
Sync'd staging from prod and fixing the xcode version so we can actually converge things.

### DIFF
--- a/data/roles/gecko_t_osx_1100_m1_loaner.yaml
+++ b/data/roles/gecko_t_osx_1100_m1_loaner.yaml
@@ -9,10 +9,10 @@ worker_metadata:
   provisionerId: "%{lookup('gw.provisioner_id')}"
 
 worker:
-  taskcluster_version: 40.0.3
+  taskcluster_version: 54.4.1
   provider_type: standalone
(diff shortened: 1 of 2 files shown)


aerickson (dismissed):
lgtm :)

data/roles/gecko_t_osx_1100_m1_loaner.yaml
```diff
@@ -9,10 +9,10 @@ worker_metadata:
   provisionerId: "%{lookup('gw.provisioner_id')}"
 
 worker:
-  taskcluster_version: 40.0.3
+  taskcluster_version: 54.4.1
```
rcurranmoz, line 12: Can you bump this to the version we deployed most recently (55.3.0)?

rcurranmoz (changes requested):
Bump TC v to 55.3.0 if possible

data/roles/gecko_t_osx_1100_m1_loaner.yaml
MasterWayZ: Done

rcurranmoz (approved):
lgtm

PR #562, opened 2023-10-03
Update robustcheckout mercurial extension and remove bundleclone

jcristau:
It seems unfortunate that there's 3 copies of robustcheckout.py (plus at least one more in the cloud_windows branch, as far as I can tell), hopefully that can be consolidated.

My main reason for updating robustcheckout is to pick up the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1481291

PR #566, opened 2023-10-06
cci: disable slack context and orb

aerickson:
We're not sending slack messages any longer, disable use of the orb and context.

The use of the slack context is making Mac jobs not run for PRs from external people (Releng, see #562).


rcurranmoz (approved):
lgtm

PR #580, opened 2023-11-08
Puppet-ized Safari Web Driver enabler

rcurranmoz:
this is already working well in SimpleMDM but I also want to see how it would look with puppet


aerickson (approved):
lgtm

PR #592, opened 2024-01-02
Cleanup mgoosens access per RELOPS-785

rcurranmoz:


aerickson:
If we delete the lines the user will still exist (because puppet doesn't know about it). We should disable the user. I think I did something similar for bclary.

aerickson (approved):

PR #598, opened 2024-01-10
Bug 1868558 - Distrust old CoT keys

JohanLorenzo:
Depends on https://github.com/mozilla-releng/scriptworker-scripts/pull/900 and https://github.com/mozilla-releng/scriptworker/pull/633


jcristau (approved):
IIRC there's also a mac used by adhoc-signing that's not covered here and will need a manual update?

JohanLorenzo:
Thanks for the review. That's right! We had to manually update it last week to get the then-latest scriptworker. We'll need it to it again.

PR #609, opened 2024-01-22
Handle different versions of Safari

rcurranmoz:
This update to macos_safaridriver retrofits @aerickson 's original solution for macOS Catalina to latter OS's.

This is currently not fully functional for the M1s as cltbld has an inconsistent userid (35 and 36 respectively). But the request was to get this working for the M2s.

This update enables safaridriver for the M2s and also supports Safari v16 and v17, as Apple moved the buttons in the UI to enable.


aerickson (approved):
nice. bad apple.

aerickson (approved):

PR #618, opened 2024-02-15
Python 3.11.0 and zstandard update for r8 pools per RELOPS-822

rcurranmoz:


modules/packages/manifests/python3.pp
```diff
@@ -3,30 +3,30 @@
...
-
 #    # As of puppet 7.0.0, facts.os.architecture still reports the M1 arm64 hardware as x86_64
 #    # therfore, we check the mac model instead
 #    if $facts['system_profiler']['model_identifier'] == 'Macmini9,1' {
 #        $pkg_name = "python-${version}-macos11.0.pkg"
 #    } else {
 #        $pkg_name = "python-${version}-macosx10.9.pkg"
 #    }
+  # Unsure why this works
+  $short_version = String($version)[0,4]
```
aerickson, line 16: `version[0..3]` gives what we're expecting, version[0,4] says start at index 0 and give me 4 chars.

https://www.dotnetperls.com/substring-ruby

`version.split('.')[0..1].join('.')` is the more future proof version (3.101.0 would give 3.101)... but we have awhile before we get there.

PR #627, opened 2024-03-08
removed defective r8s

rcurranmoz:


modules/packages/manifests/python3.pp
aerickson, line 1: let's remove the changes in this file from this pr... i think i've got a fix in https://github.com/mozilla-platform-ops/ronin_puppet/pull/628.

PR #635, opened 2024-03-14
change default uid for cltbld/jupdate saf driver

rcurranmoz:


rcurranmoz:
Previous to macOS 13, the cltbld user would be assigned the next available ID on the system during provisioning. Usually 35 or 36. Starting with macOS 13+, having a uid set too low results in Launch Agents/Login Items to function. Specifically they won't load from /Users/cltbld/Library/Launch\ Agents/

Setting the ID > 501 fixes the issue.

Additionally we hard code the UID in our Safari Driver automation. This PR addresses both issues.

PR #649, opened 2024-04-11
M2 role profile fix

rcurranmoz:


rcurranmoz:
I pushed an accidental commit out of order so I'm merging this one to resolve the issue. As you can see no files get changed.

gibbsju (approved):

PR #659, opened 2024-04-24
circleci: fix mac executors

aerickson:
We're experiencing rolling blackouts as part of https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718. See https://mozilla-hub.atlassian.net/browse/RELOPS-936.

Switch off the old mac instance types (by defining a current instance type).

Also:
- add tentative support for the OS X version CircleCI is running (21)
- upgrade linux instance types
- upgrade orbs


rcurranmoz (approved):
Looks good. Thank you

PR #663, opened 2024-04-30
Add Safari Tech update script/permissions for cltbld

rcurranmoz:
This enables `cltbld` to run the `softwareupdate` in order to install the latest Safari Technology Preview before running the test.

https://mozilla-hub.atlassian.net/browse/RELOPS-866

`sudoers` after applying:
```
## concatenated files:
cltbld ALL=(root) NOPASSWD: /sbin/reboot
cltbld ALL=(root) NOPASSWD: /usr/local/bin/install_safari_softwareupdates.py
```

### DIFF
--- a/modules/roles_profiles/manifests/profiles/cltbld_user.pp
+++ b/modules/roles_profiles/manifests/profiles/cltbld_user.pp
@@ -83,7 +83,7 @@ class roles_profiles::profiles::cltbld_user {
         # require => User['cltbld'],
       }
 
-      $sudo_commands = ['/sbin/reboot']
+      $sudo_commands = ['/sbin/reboot', '/usr/sbin/softwareupdate']
       $sudo_commands.each |String $command| {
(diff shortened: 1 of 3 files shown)


modules/roles_profiles/manifests/profiles/cltbld_user.pp
```diff
@@ -83,7 +83,7 @@
         # require => User['cltbld'],
       }
 
-      $sudo_commands = ['/sbin/reboot']
+      $sudo_commands = ['/sbin/reboot', '/usr/sbin/softwareupdate']
```
aerickson, line 86: We don't want to allow this command, but rather the python script I wrote that only allows updating 'safari' packages.

rcurranmoz: Oh got it. For some reason I thought that would go in the test itself. Will fix

aerickson (approved):
lgtm

PR #668, opened 2024-05-14
Bug 1896540 - update windows maintenance service registry entry for n…

jcristau:
…ew code signing cert

We haven't used Thawte as a CA in forever, so replace it with the current issuer.


bhearsum (approved):

jcristau:
Thanks!

PR #634, opened 2024-03-14
moonshot linux: py311, hg, and tc component update

aerickson:
See https://mozilla-hub.atlassian.net/browse/RELOPS-821 and https://mozilla-hub.atlassian.net/browse/RELOPS-846.

- Install python3.11 (not symlinked to python3 though).
- TC component upgrade
- mercurial upgrade

Also includes:
- kitchen_docker fixing 
- CircleCI image updates

PR #679, opened 2024-06-26
moonshot linux py311 and tc update plus 1804 hg upgrade

aerickson:
- **talos moonshot: use v61 of tc components**
- **kitchen docker: puppet-agent install fix**
- **install py3.11 and set as default**
- **py3 test tweaks**
- **wip**
- **alternatives fixing**
- **disable alternative making python3 call python3.11**
- **fix test**
- **place 1804_hg_upgrade changes on top of moonshot_linux_py311_and_tc_update**
- **re-enable mac py2zstandard, disable linux py2zstandard**
- **run-puppet file can load settings from a non-vc path**


aerickson:
Merged into #634.

PR #691, opened 2024-07-17
RELOPS-1012: Safari driver update

rcurranmoz:
Problem: we hard code the `UID` of the `cltbld` user when executing the Safari Driver automation. Sometimes the `UID` can vary, so having it dynamically pulled will help with puppet applys


aerickson (approved):
nice!

PR #693, opened 2024-07-18
Change cltbld account provisioning

rcurranmoz:
Historically we've been setting the password for the `cltbld` account using salt/iterations/hashing. The end result was a secure password but also one that was unknown to us in plaintext.

In an effort to shift to Generic Worker multiuser, we will need to hard code the credentials for `cltbld` in` /opt/worker/current-task-user.json` and `/opt/worker/next-task-user.json.`

This PR sets a plaintext password stored in Vault.

@aerickson I don't think this will break anything on your end, but can you check line 67 and just ensure there will be no conflict. I changed `cltbld_user.password` in Vault, but only in `roles/gecko_t_osx_1015_r8_staging/vault_secrets::cltbld_user
`

Thanks


rcurranmoz:
We went from using the `user` provisioner in puppet to the `exec`. Could that be why?

PR #695, opened 2024-07-22
Use chocolatey for google chrome

jwmoss:


modules/win_scheduled_tasks/files/maintainsystem-reftester.ps1
```diff
@@ -291,6 +291,35 @@ function Get-LoggedInUser {
     @(((query user) -replace '\s{20,39}', ',,') -replace '\s{2,}', ',' | ConvertFrom-Csv)
 }
 
+function Get-LatestGoogleChrome {
```
markcor, line 294: This code should be executed by Puppet in `win_packages::chrome` .  It can be a script in the win_package file dir and then called by an exec resource before the package is installed. 

Not for this PR, we should consolidate to one maintain system script for hardware.

markcor, line 294: This code should be executed by Puppet in `win_packages::chrome` .  It can be a script in the win_package file dir and then called by an exec resource before the package is installed. 

Not for this PR, we should consolidate to one maintain system script for hardware.

markcor (changes requested):

markcor (approved):

PR #696, opened 2024-07-23
RELOPS-838 Add directory_cleaner module to cleanup GW

rcurranmoz:
For the last few weeks we've seen `/opt/worker/downloads` and `/opt/worker/cache` on the r8s not getting cleaned up by generic worker, causing subsequent test failures due to insufficient amounts of space.

This runs the tool @aerickson created before the machine reboots between tasks.


aerickson (approved):
lgtm

PR #698, opened 2024-07-25
added symlink fix for hg

rcurranmoz:
`hg` needs to be seen at `/usr/local/bin/hg` for tests to pass


jcristau:
Cross-reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1909904

PR #705, opened 2024-07-31
sbom generator for macOS

rcurranmoz:
first draft. outputs to `/var/sbom/sbom.md` for now

### DIFF
--- a/modules/macos_sbom/files/generate_sbom.py
+++ b/modules/macos_sbom/files/generate_sbom.py
@@ -16,11 +16,26 @@ def log_message(message):
     with open(LOG_FILE, 'a') as log_file:
         log_file.write(message + '\n')
 
+def get_macos_version():
+    try:
+        log_message("Fetching macOS version...")
+        result = subprocess.run(['/usr/sbin/system_profiler', 'SPSoftwareDataType'], capture_output=True, text=True, check=True, timeout=60)
+        for line in result.stdout.split('\n'):
+            if 'System Version' in line:
+                version = line.split(':')[-1].strip()
+                log_message(f"macOS version found: {version}")
+                return version
+    except subprocess.CalledProcessError as e:
+        log_message(f"Error while fetching macOS version: {e}")
+    except subprocess.TimeoutExpired:
+        log_message("Timeout while fetching macOS version")
+    return "Unknown version"
+
 def get_installed_software():
     try:
         log_message("Fetching installed software...")
         result = subprocess.run(['/usr/sbin/system_profiler', 'SPApplicationsDataType', '-json'], capture_output=True, text=True, check=True, timeout=60)
-        log_message(f"Installed software fetched successfully: {result}")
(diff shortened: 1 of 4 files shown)


modules/macos_sbom/manifests/init.pp
```diff
@@ -0,0 +1,38 @@
+class macos_sbom (
+  Boolean $enabled = true,
+) {
+  # Ensure Python3 is installed
```
aerickson, line 4: nit: remove if not needed

aerickson (commented):
We're using Markdown format for linux and windows SBOMs. It might be nice to all use the same format.

On linux/monopacker, I'm writing to `/etc/SBOM.md` if you want to standardize for POSIX systems. Open to changing that if you think it's a bad path.

rcurranmoz:
> We're using Markdown format for linux and windows SBOMs. It might be nice to all use the same format.
> 
> On linux/monopacker, I'm writing to `/etc/SBOM.md` if you want to standardize for POSIX systems. Open to changing that if you think it's a bad path.

Na I did it in `html` at first out of my own ignorance as to how a browser processes markdown. Aka the` .md` didn't look pretty opening directly in a browser but TIL something else must be happening to make `.md` files on GH look "pretty" for example lol

Fixed

modules/macos_sbom/files/generate_sbom.py
```diff
@@ -0,0 +1,134 @@
...
+        log_message(f"Error while fetching pip packages: {e}")
+        return []
+    except subprocess.TimeoutExpired:
+        log_message("Timeout while fetching pip packages")
+        return []
+
+def generate_sbom(applications, binaries, pip_packages):
+    sbom = {
+        "sbom_version": "1.0",
+        "macOS_version": "14",
```
aerickson, line 80: We could use `system_profiler SPSoftwareDataType | grep 'System Version'` vs a static version.

```bash
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 14.5 (23F79)
      Kernel Version: Darwin 23.5.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Computer Name: powderdry
      User Name: Andrew Erickson (aerickson)
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 40 days, 20 hours, 47 minutes
```

aerickson:
It would be sweet if the ronin-puppet git sha1 that was used to converge was in the SBOM. 

the way we do a git dir per 'environment' is a bit tricky. we could maintain a 'latest' symlink that points to the last 'environment' used to converge, then the script could cd to that dir and run git show. or we could just grab the production environment's git sha and understand that the sbom will be wrong on our non-production nodes.

rcurranmoz:
@aerickson I might need you to ELIF

aerickson:
LGTM. Let's land this as is for now and we can continue iterating after.

aerickson (approved):

PR #715, opened 2024-08-08
xcode cli tools install

rcurranmoz:
Switches to dynamically installing xcode command line tools (vs uploading each minor version to s3)

Curious if you see any downsides here @aerickson

### DIFF
--- a/data/roles/gecko_t_osx_1400_m2_staging.yaml
+++ b/data/roles/gecko_t_osx_1400_m2_staging.yaml
@@ -33,7 +33,7 @@ packages_classes:
   - tooltool
   - virtualenv
   - wget
-  - xcode_cmd_line_tools
+  #- xcode_cmd_line_tools
   - zstandard
   - telegraf
   - virt_audio_s3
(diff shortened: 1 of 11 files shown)


aerickson (approved):
lgtm

aerickson:
Ah, automated tests may have caught something? 

> Error: Could not find resource 'Class[Macos_xcode_tools]' in parameter 'require' (file: /tmp/kitchen/modules/packages/manifests/mercurial.pp, line: 33) on node static.38.23.41.44.cyberlynk.net

rcurranmoz:
@aerickson i did see that but wasn't sure why. The class seems valid. The only thing I noticed is that the error seems to indicate test kitchen is checking for a:

`'Class[Macos_xcode_tools]'`

vs 

`'Class[macos_xcode_tools]'`

but I imagine that isn't it?

aerickson:
I think it was just quoting (see https://github.com/mozilla-platform-ops/ronin_puppet/pull/715/commits/ef8ff318b2f39232ced63e7ab05a4d1ed47bb16d). Rebased and pushed a fix, but still a conflict you have to resolve.

PR #716, opened 2024-08-20
[RELOPS-986] Added generic worker process checker

rcurranmoz:
https://mozilla-hub.atlassian.net/browse/RELOPS-986

macOS workers (especially performance workers for some reason) routinely stop taking tasks after 24-48 hours. This module lands a script that gets run every 30 minutes to see if the Generic Worker process is running and if it's not, reboots the host. Rebooting the host always "fixes" the issue.

This solution was manually tested prior to Mozweek and was successful.

The Taskcluster team suggests we switch to multiuser generic worker in hope of resolving this issue. While that might still be a priority, since it is a much larger lift, this PR is the short term solution.

### DIFF
--- a/modules/macos_gw_checker/files/gw_checker.sh
+++ b/modules/macos_gw_checker/files/gw_checker.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Check if the semaphore file exists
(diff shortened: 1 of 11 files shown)


aerickson (approved):
lgtm

PR #727, opened 2024-08-30
TCC permissions spec test

rcurranmoz:


test/integration/gecko_t_osx_1015_r8/serverspec/tcc_perms_spec.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'spec_helper'
+require 'sqlite3'
```
rcurranmoz, line 2: Is there a way we can add the `sqlite3` package @aerickson ?

aerickson: Hmm, yeah, I think we could probably get it installed in the Ruby environment that Puppet uses. 

You could also try to use the `sqlite` binary to run the query to avoid having to install the Ruby gem.

PR #729, opened 2024-08-30
Wifi bug fix

rcurranmoz:
macOS systems seem to have turned wifi back on (via possibly a software update/apple being annoying) and the existing puppet code didn't remediate the re-enablement.

Now it does


aerickson (approved):
lgtm

PR #732, opened 2024-08-30
bluetooth spec test

rcurranmoz:


rcurranmoz:
"The failure indicates that the test expected the value of BluetoothAutoSeekKeyboard to be 1, but it was actually 0. This suggests that the current state of the system does not match the expected configuration."

PR #736, opened 2024-08-30
packages install spec

rcurranmoz:


rcurranmoz:
The failure indicates that the installed version of Python 3 is 3.11.3 rather than the expected 3.11.0.

aerickson:
> The failure indicates that the installed version of Python 3 is 3.11.3 rather than the expected 3.11.0. Thoughts @aerickson

Strange... we install 3.11.0 via a Homebrew pkg right? I guess the CircleCI image could already have Python 3.11.3 installed? The 3.11.3 path must closer to the start of PATH and therefore wins.

We could relax the requirement (only check for 3.11) or we could specify the full path to the binary we want to check (then it should be the correct version).

rcurranmoz:
> > The failure indicates that the installed version of Python 3 is 3.11.3 rather than the expected 3.11.0. Thoughts @aerickson
> 
> Strange... we install 3.11.0 via a Homebrew pkg right? I guess the CircleCI image could already have Python 3.11.3 installed? The 3.11.3 path must closer to the start of PATH and therefore wins.
> 
> We could relax the requirement (only check for 3.11) or we could specify the full path to the binary we want to check (then it should be the correct version).

You know, that may have been how it was done prior to our latest rollout. As of the latest rollout we've been using the Python 3 pkg from https://www.python.org/ , which in hindsight explains the issue I ran into with the new `hg` living in a different place.

But the `pkg` we deploy is in fact 3.11.0

PR #744, opened 2024-09-13
Refactor signing_worker requirements utility + bump scriptworker revisions

ahal:
@hneiva the old docker based script wasn't working for me and I was having trouble figuring out what was going wrong. I think this new script is much simpler and runs really fast. It works because of the `--universal` flag which means the requirements.txt is cross platform and cross python version. We can limit the generation to macos if desired however.


hneiva (changes requested):
vpn workers are missing the update in `data/common.yaml`

hneiva (approved):
LGTM. Thanks!

PR #747, opened 2024-09-16
[RELOPS-773] New mac signer pool/deprecating Deploy Studio

rcurranmoz:
We're still in the early stages of [understanding](https://docs.google.com/document/d/1pgIK4TbeqJtRlt1UiE4C7o2e37I2DttUrxEcN72d5Pg/edit) how the mac signers are configured, but I think we are getting close. The purpose of this PR is to setup a new pool initially for testing, which will then convert to prod.

I *think* the code below will need to merge before the auto puppeting of the new signer host takes place

We can change the hostnames at a later time but the short term goal is to just see if we can get a signer provisioned on new hardware manually.

Please sanity check me @ahal @aerickson and/or @hneiva as I know the signers are sensitive. Thanks all

### DIFF
--- a/data/roles/mac_v4_signing_ff_prod.yaml
+++ b/data/roles/mac_v4_signing_ff_prod.yaml
@@ -1,0 +1,13 @@
+---
+
+# Package class parameters
+packages::vault::version: 1.7.3
+
+# Override secrets with vault secrets
+
+papertrail:
+  host: "%{lookup('vault_secrets::papertrail.data.host')}"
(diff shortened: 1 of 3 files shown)


aerickson:
LGTM

> I think the code below will need to merge before the auto puppeting of the new signer host takes place

Yeah, I think so unless you want to point them at a branch other than master.

hneiva:
FWIW if you push this to a branch, we can test it manually on the machine by pointing the run-puppet script to the branch

aerickson (approved):

PR #760, opened 2024-10-14
moonshot linux chrome fix 101424

aerickson:
- **1804: install older chrome version from s3**
- **re-enable the spec test on 1804**


rcurranmoz (approved):
lgtm!

PR #770, opened 2024-10-25
[RELOPS-1131] Add two mdc m2 workers to gecko-3-b-osx-arm64

rcurranmoz:
To replace the AWS workers


aerickson (approved):
LGTM. 

Annoying there's no easy way to have two different worker groups, I guess we could make another role, but not worth it (hopefully we never have to converge another aws mac again).

PR #774, opened 2024-11-15
[RELOPS-1145] Add new path to directory_cleaner

rcurranmoz:
https://mozilla-hub.atlassian.net/browse/RELOPS-1145

PR #775, opened 2024-11-15
Clean up org.mozilla.* in /var/folders

rcurranmoz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1927416

Plan is to deploy this to the mac workers after work week next week

### DIFF
--- a/modules/worker_runner/templates/worker-runner.sh.erb
+++ b/modules/worker_runner/templates/worker-runner.sh.erb
@@ -20,6 +20,7 @@ rm <%= @data_dir %>/tasks-resolved-count.txt
 
 # Clear up previous task dirs
 rm -rf /opt/worker/tasks/task_*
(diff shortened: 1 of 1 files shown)


modules/worker_runner/templates/worker-runner.sh.erb
```diff
@@ -20,6 +20,7 @@ rm <%= @data_dir %>/tasks-resolved-count.txt
 
 # Clear up previous task dirs
 rm -rf /opt/worker/tasks/task_*
+find /private/var/folders/ -path '*/C/org.mozilla.*' -type d -o -path '*/C/org.mozilla.*/*' -type d -exec rm -rf {} +
```
whimboo, line 23: Should we as well clean up the `Download` folder? Maybe a new section for both makes sense given that both are not related to the `task dirs`.

rcurranmoz: That's in this PR handled by a different tool https://github.com/mozilla-platform-ops/ronin_puppet/pull/774

PR #784, opened 2024-12-16
[RELOPS-1164] Deploy watchman (hg fsmonitor)

rcurranmoz:
https://mozilla-hub.atlassian.net/browse/RELOPS-1164

https://github.com/facebook/watchman/releases/tag/v2023.05.01.00

`watchman.pkg` drops 
```
watchman 
watchmanctl
```

into `/usr/local/bin` and

```
libcrypto.1.1.dylib
libevent-2.1.7.dylib
libgflags.2.2.dylib
libglog.0.dylib
liblz4.1.dylib
libpcre2-8.0.dylib
libsodium.23.dylib
libssl.1.1.dylib
libz.1.dylib
libzstd.1.dylib
```

into `/usr/local/lib/`

These binaries should work for now but its worth noting that this the last binary release for macOS. Other install methods are available via `homebrew`, but that presents challenges (for one we dont install `homebrew`) and my initial testing via the `brew pkg` method were unsuccessful

### DIFF
--- a/inventory.d/macmini-r8.yaml
+++ b/inventory.d/macmini-r8.yaml
@@ -218,6 +218,8 @@ groups:
 
   - name: mozillavpn-b-1-osx
     targets:
+    # New Mac15 VPN worker
+      - macmini-r8-102.test.releng.mdc1.mozilla.com
       - macmini-r8-179.test.releng.mdc1.mozilla.com
       - macmini-r8-182.test.releng.mdc1.mozilla.com
       - macmini-r8-183.test.releng.mdc1.mozilla.com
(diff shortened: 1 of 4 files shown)


aerickson:
Hmm, do we need to update hgrc to enable the plugin? I think we might.

rcurranmoz:
> Hmm, do we need to update hgrc to enable the plugin? I think we might.

do you think that would cause these failures?

https://treeherder.mozilla.org/jobs?repo=try&revision=edbd01fa4042eb0bb982b2542575a5269f647f10

aerickson:
I found `could not find user worker; specify a valid user with --user` in some of the logs. The others look like intermittent failures.

Not sure why it would be looking for that user... test misconfiguration?

Let's rerun with the hgrc placed and see what happens.

aerickson (dismissed):
LGTM. Let's fix the one test failure before landing.

PR #788, opened 2025-01-02
add cltbld check to saf driver

rcurranmoz:
when provisioning a host from scratch this automation as is will run under the logged in user `relops` but we need to ensure it only runs as `cltbld`


aerickson (approved):
nice

PR #789, opened 2025-01-02
fix cli tools checker

rcurranmoz:
on mac 14 this is failing to run because `/Library/Developer/CommandLineTools/usr` exists, despite the command line tools not being installed


aerickson (approved):
r+

PR #808, opened 2025-02-03
refactor: :coffin: Remove common_tools.pp with unneeded tooling

jwmoss:
This removes procmon, gpg4win and sublimetext from all windows workers.

### DIFF
--- a/modules/roles_profiles/manifests/roles/win11642009hwref.pp
+++ b/modules/roles_profiles/manifests/roles/win11642009hwref.pp
@@ -22,6 +22,7 @@ class roles_profiles::roles::win11642009hwref {
 
   # Adminstration
   include roles_profiles::profiles::logging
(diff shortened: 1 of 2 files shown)


modules/roles_profiles/manifests/roles/win11a6424h2azurebuilder.pp
```diff
@@ -32,11 +32,11 @@
 
       # Adminstration
       include roles_profiles::profiles::logging
-      include roles_profiles::profiles::common_tools
 
       # Worker
       include roles_profiles::profiles::google_auth
       include roles_profiles::profiles::git
+      include roles_profiles::profiles::gpg4win
```
markcor, line 39: We might be able to remove this gpg42in class entirely.

markcor (approved):

PR #824, opened 2025-02-27
RELOPS-591 Auto Puppetize Macs

rcurranmoz:
https://mozilla-hub.atlassian.net/browse/RELOPS-591

### DIFF
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -7,6 +7,10 @@ defaults:
 hierarchy:
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
+
+  - name: "local"
+    path: "/var/root/vault.yaml"
+
   - name: "Per-role data"
     path: "roles/%{facts.puppet_role}.yaml"
 
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -8,9 +8,6 @@ hierarchy:
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
 
-  - name: "local"
-    path: "/var/root/vault.yaml"
-
   - name: "Per-role data"
     path: "roles/%{facts.puppet_role}.yaml"
 
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -8,6 +8,9 @@ hierarchy:
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
 
+  - name: "local"
+    path: "/var/root/vault.yaml"
+
   - name: "Per-role data"
     path: "roles/%{facts.puppet_role}.yaml"
 
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -7,7 +7,6 @@ defaults:
 hierarchy:
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
-
   - name: "Per-role data"
     path: "roles/%{facts.puppet_role}.yaml"
 
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -5,6 +5,10 @@ defaults:
   datadir: data
 
 hierarchy:
+
+  - name: "Secrets generated from Vault"
+    path: "secrets/vault.yaml"
+
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
   - name: "Per-role data"
@@ -16,9 +20,6 @@ hierarchy:
 
   - name: "Per-OS defaults"
     path: "os/%{facts.os.family}.yaml"
-
-  - name: "Secrets generated from Vault"
-    path: "secrets/vault.yaml"
 
   - name: "Common data to all"
     path: "common.yaml"
--- a/modules/macos_auto_puppet/files/auto-puppet.sh
+++ b/modules/macos_auto_puppet/files/auto-puppet.sh
@@ -14,7 +14,7 @@ PUPPET_ROLE_FILE="/etc/puppet_role"
 PUPPET_BIN="/opt/puppetlabs/bin/puppet"
 FACTER_BIN="/opt/puppetlabs/bin/facter"
 GIT_REPO_URL="https://github.com/mozilla-platform-ops/ronin_puppet.git"
-GIT_BRANCH="master"
+GIT_BRANCH="auto_puppet_v2"
 
 # Override defaults with values from /etc/puppet/ronin_settings if the file exists
 if [ -f "/etc/puppet/ronin_settings" ]; then
@@ -58,9 +58,21 @@ if [ ! -x "$FACTER_BIN" ]; then
     fail "Facter is missing or not executable."
 fi
 
-# Remove existing Puppet repository and re-clone it
-rm -rf "$LOCAL_PUPPET_REPO"
-git clone --branch "$GIT_BRANCH" "$GIT_REPO_URL" "$LOCAL_PUPPET_REPO" || fail "Failed to clone Puppet repository"
+# Clone or update Puppet repository
+if [ -d "$LOCAL_PUPPET_REPO/.git" ]; then
+    echo "Updating existing Puppet repository..."
+    cd "$LOCAL_PUPPET_REPO" || fail "Failed to change directory to Puppet repository"
+
+    # Reset any local changes and pull the latest changes
+    git reset --hard HEAD || fail "Failed to reset repository"
+    git clean -fd || fail "Failed to clean untracked files"
+    git fetch --all || fail "Failed to fetch latest changes"
+    git checkout "$GIT_BRANCH" || fail "Failed to checkout branch $GIT_BRANCH"
+    git pull origin "$GIT_BRANCH" || fail "Failed to pull latest changes"
+else
+    echo "Cloning Puppet repository..."
+    git clone --branch "$GIT_BRANCH" "$GIT_REPO_URL" "$LOCAL_PUPPET_REPO" || fail "Failed to clone Puppet repository"
+fi
 
 # Ensure Puppet Repository Exists
 get_puppet_repo() {
--- a/modules/macos_auto_puppet/files/auto-puppet.sh
+++ b/modules/macos_auto_puppet/files/auto-puppet.sh
@@ -16,6 +16,21 @@ FACTER_BIN="/opt/puppetlabs/bin/facter"
 GIT_REPO_URL="https://github.com/mozilla-platform-ops/ronin_puppet.git"
 GIT_BRANCH="auto_puppet_v2"
 
+# Override defaults with values from /etc/puppet/ronin_settings if the file exists
+if [ -f "/etc/puppet/ronin_settings" ]; then
+    echo "Loading settings from /etc/puppet/ronin_settings..."
+    source /etc/puppet/ronin_settings
+
+    # Explicitly export the variables to ensure they override the defaults
+    export GIT_REPO_URL="${PUPPET_REPO:-$GIT_REPO_URL}"
+    export GIT_BRANCH="${PUPPET_BRANCH:-$GIT_BRANCH}"
+    export PUPPET_MAIL="${PUPPET_MAIL:-}"
+    export WORKER_TYPE_OVERRIDE="${WORKER_TYPE_OVERRIDE:-}"
+fi
+
+echo "Using Puppet Repo: $GIT_REPO_URL"
+echo "Using Branch: $GIT_BRANCH"
+
 # Vault configuration
 export VAULT_ADDR="http://127.0.0.1:8200"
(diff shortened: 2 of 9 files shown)


modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,121 @@
...
+    fail "Puppet is missing or not executable."
+fi
+
+if [ ! -x "$FACTER_BIN" ]; then
+    fail "Facter is missing or not executable."
+fi
+
+# Remove existing Puppet repository and re-clone it
+rm -rf "$LOCAL_PUPPET_REPO"
+git clone --branch "$GIT_BRANCH" "$GIT_REPO_URL" "$LOCAL_PUPPET_REPO" || fail "Failed to clone Puppet repository"
```
aerickson, line 48: A bit wasteful, but i guess it avoids issues with the checkout dir (PUPPET_ENV) that I had to deal with on linux.

hiera.yaml
```diff
@@ -7,6 +7,10 @@ defaults:
 hierarchy:
     # It is acceptable for dot notation facts (eg. %{facts.some_fact}) to not exist
     # since they will simply return an empty string
+
```
aerickson, line 10: linux gets by without adding this by copying it into place. 

https://github.com/mozilla-platform-ops/ronin_puppet/blob/eff837c22e94dc7e7c230936d5bb192d971a39a2/modules/puppet/templates/puppet-ubuntu-run-puppet.sh.erb#L147

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,121 @@
...
+export LANG=en_US.UTF-8
+export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/puppetlabs/bin"
+
+# Puppet configuration
+LOCAL_PUPPET_REPO="/Users/relops/Desktop/puppet/ronin_puppet"
+PUPPET_ROLE_FILE="/etc/puppet_role"
+PUPPET_BIN="/opt/puppetlabs/bin/puppet"
+FACTER_BIN="/opt/puppetlabs/bin/facter"
+GIT_REPO_URL="https://github.com/mozilla-platform-ops/ronin_puppet.git"
+GIT_BRANCH="auto_puppet_v2"
```
aerickson, line 17: make sure to not land with this

modules/macos_auto_puppet/manifests/init.pp
```diff
@@ -0,0 +1,65 @@
+class macos_auto_puppet (
+  Boolean $enabled = true,
+  String $puppet_role = $facts['puppet_role'], # Dynamically pull from Facter
```
aerickson, line 3: This facter dance seems odd. Does linux get around this with https://github.com/mozilla-platform-ops/ronin_puppet/blob/eff837c22e94dc7e7c230936d5bb192d971a39a2/modules/puppet/templates/puppet-ubuntu-run-puppet.sh.erb#L152?

hiera.yaml
rcurranmoz: Thank you. Removed (and still working)

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,121 @@
...
+export VAULT_ADDR="http://127.0.0.1:8200"
+VAULT_TOKEN="$(cat /etc/vault_token 2>/dev/null)"
+export VAULT_TOKEN
+
+# Fail function
+fail() {
+    echo "${@}"
+    exit 1
+}
+
```
aerickson, line 29: I'd include the linux override stuff... it's super handy and I forsee us wanting it on Macs soon once this lands.

rcurranmoz: Great idea. I will

hiera.yaml
rcurranmoz: idk why it breaks when i remove this but i moved the logic into the mac `auto-puppet.sh`. works

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,164 @@
...
+# Ensure Puppet Repository Exists
+get_puppet_repo() {
+    if [ ! -d "$LOCAL_PUPPET_REPO" ]; then
+        fail "Local Puppet repository not found at $LOCAL_PUPPET_REPO"
+    fi
+    cd "$LOCAL_PUPPET_REPO" || fail "Failed to change directory to $LOCAL_PUPPET_REPO"
+
+    # Inject Hiera Configuration
+    sudo tee "$LOCAL_PUPPET_REPO/hiera.yaml" > /dev/null << 'EOF'
```
aerickson, line 73: On linux, we copy the vault.yaml to the expected location. 

Dropping in a new hiera.conf works, but you have to kind of keep this in sync with the main one.

modules/macos_auto_puppet/manifests/init.pp
aerickson: OK, so yeah, Bolt sets puppet_role for you... so now that we've eliminated it, it's not set. 

I think we can set the Puppet fact just on the CLI when we run puppet in auto-puppet.sh (which can read it from the role file) with:

```
ROLE=$(cat /etc/puppet_role)
FACTER_puppet_role=${ROLE} puppet apply manifest.pp
```

I guess having a durable fact is nice... are we ever going to run Puppet on these outside of auto-puppet.sh thought?

rcurranmoz: Not that I can foresee. However, if we implement it, it could open up opportunities for further tweaking and optimization as usage evolves over time.

aerickson: I still don't get how the `macos_auto_puppet` class works. It takes a fact (that's not set or it would be working without this code) and then sets the same fact (so it works)? 

Where is it getting the value from initially?

hiera.yaml
```diff
@@ -7,6 +7,10 @@ defaults:
...
+  - name: "local"
+    path: "/var/root/vault.yaml"
```
aerickson: If we copy `/var/root/vault.yaml` to the location on line 21/25 below in auto-puppet.sh (relative to the repo checkout) then we don't need to add anything.

aerickson: Ah this was looking at an old revision. Do you just place it during setup at this location?

aerickson: Ah, I see we drop an updated hiera.conf in auto-puppet.sh.

modules/macos_auto_puppet/files/auto-puppet.sh
rcurranmoz: True but I've never had to change `hiera.yaml` in the last 18 months, so i think its going to be relatively burden-free

aerickson: Yeah, I guess it comes down to being a single line (copying the file) or several lines (placing a hiera conf), I'd go with copying the file.

On line 102 below you're already doing the copy... is it just not making it into the correct place?

modules/macos_auto_puppet/files/auto-puppet.sh
aerickson: We were just talking about DC network usage and the ticket about throttling came up. 

Github has been pretty good about not throttling us, but if they did start this would become a problem. Since it's in the critical flow of the worker, no tasks would be able to run.

The linux script (I think) has solved the issue with updating existing clients.

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,176 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Purpose: Bootstrap a macOS 14+ host from post-install (or image) to a complete Puppet run.
+
+export LANG=en_US.UTF-8
+export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/puppetlabs/bin"
+
+# Puppet configuration
+LOCAL_PUPPET_REPO="/Users/relops/Desktop/puppet/ronin_puppet"
```
aerickson, line 12: The issue we ran into on linux is that when people change the GIT_REPO_URL in this script (or in the override), the git fetch or checkout below fail because it's not a fast-forward update or something. I fixed this for most cases on linux by using the GH username from the URL in the repo path (https://github.com/mozilla-platform-ops/ronin_puppet/blob/331a7b0fbfb8c3a1af94e9d14b564d8fa853cbcc/modules/puppet/templates/puppet-ubuntu-run-puppet.sh.erb#L37). I think there are other ways around it like telling pull/checkout to do non-fast-forwards or something.

aerickson: Nice. LGTM.

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,180 @@
...
+PUPPET_ROLE_FILE="/etc/puppet_role"
+PUPPET_BIN="/opt/puppetlabs/bin/puppet"
+FACTER_BIN="/opt/puppetlabs/bin/facter"
+GIT_REPO_URL="https://github.com/mozilla-platform-ops/ronin_puppet.git"
+GIT_BRANCH="auto_puppet_v2"
+
+# Override defaults with values from /etc/puppet/ronin_settings if the file exists
```
aerickson, line 19: Let's reorganize this file and put non-functions after the functions. It's a bit tricky to read now.

So, this order:
- variable definitions (like we have currently)
- functions
- main/non-functions

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,180 @@
...
+            echo "Already up-to-date. No changes needed."
+        fi
+    fi
+else
+    echo "Cloning fresh Puppet repository..."
+    git clone --branch "$GIT_BRANCH" "$GIT_REPO_URL" "$LOCAL_PUPPET_REPO" || fail "Failed to clone Puppet repository"
+fi
+
+# Ensure Puppet Repository Exists
+get_puppet_repo() {
```
aerickson, line 88: Let's rename this function or make it do what it says.

It does seem like a good idea to put the checkout in a function as it's a large block of code.

hiera.yaml
```diff
@@ -5,6 +5,10 @@ defaults:
   datadir: data
 
 hierarchy:
+
+  - name: "Secrets generated from Vault"
```
aerickson, line 9: Much more palatable... I'd still like to understand why this makes it work.

modules/macos_auto_puppet/manifests/init.pp
```diff
@@ -0,0 +1,51 @@
...
+  # Set Puppet role dynamically
+  file { '/etc/facter/facts.d/puppet_role.txt':
+    ensure  => file,
+    content => "puppet_role=${puppet_role}\n",
+    owner   => 'root',
+    group   => 'wheel',
+    mode    => '0644',
+  }
+
+  # Ensure /etc/puppet_role exists and matches Facter
```
aerickson, line 34: Let's have puppet leave this file alone. 

- We manage it manually on linux (and in MDM on Mac).
- It helps maintain sanity when debugging if only we alter it.

modules/macos_auto_puppet/files/auto-puppet.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Purpose: Bootstrap a macOS 14+ host from post-install (or image) to a complete Puppet run.
+
```
aerickson, line 7: let's `set -e` for sure

rcurranmoz:
It's probably best to just close this and let the next person figure it out. Closing for now

modules/macos_auto_puppet/files/auto-puppet.sh
rcurranmoz: Pushed a new commit. Thank you

modules/macos_auto_puppet/files/auto-puppet.sh
rcurranmoz: Pushed a new commit. Thank you

modules/macos_auto_puppet/manifests/init.pp
rcurranmoz: Sounds good. So basically land `/etc/puppet_role` with a pkg via mdm-- like I do with the signers?

rcurranmoz:
Closing this PR and reopening a fresh one — GitHub didn’t pick up latest commits after a force push + reopen.

aerickson:
> Closing this PR and reopening a fresh one — GitHub didn’t pick up latest commits after a force push + reopen.

Hmm, let's zoom and I bet we can get this working.

PR #832, opened 2025-03-19
fix(signing_worker): Bump scriptworker and scriptworker-scripts to la…

ahal:
…test revisions


ahal:
This is needed to pick up the fix to add maple as a trusted source tree.

Looks like we haven't done this in awhile.

bhearsum (approved):

gabrielBusta (approved):

PR #841, opened 2025-04-09
[RELOPS-1426] Remove decom workers/add new worker

rcurranmoz:
Removed ec2 workers as they have been decommissioned. Added additional worker per request.


markcor (approved):

ctbfourone (approved):
I (obviously) don't know the full context but this seems legit to me! And it matches the host you mentioned in IO-3174.

PR #854, opened 2025-05-01
[RELOPS-1459] disable disableOOMProtection on windows profile

jwmoss:


modules/win_taskcluster/templates/standalone_runner.yml.epp
```diff
@@ -17,6 +17,7 @@ workerConfig:
   cachesDir: "<%= $win_taskcluster::worker_runner::cache_dir %>"
   cleanUpTaskDirs: true
   disableReboots: false
+  disableOOMProtection: true
```
rcurranmoz, line 20: this might need a comma, yea?

rcurranmoz (approved):
lgtm

modules/win_taskcluster/templates/standalone_runner.yml.epp
ctbfourone: I think because this gets rendered to YAML, where the .config.epp template looks like JSON.

PR #856, opened 2025-05-07
Handle Python dependency issues on VPN builder jobs

rcurranmoz:
VPN builders are failing jobs when this change is applied.

We usually resolve these types of Puppet failures by uploading the necessary .whl files, but that may not be practical for all the dependencies used in this job.
In this case, the missing dependency was multidict, but fixing that may just uncover additional missing packages.

@aerickson — let me know what you think.

I think its also useful to note that while this config change was added to the builders in ronin_puppet some time ago, we have not yet to date applied the config (until now) to builders

### DIFF
--- a/modules/roles_profiles/manifests/roles/mozilla_b_1_osx.pp
+++ b/modules/roles_profiles/manifests/roles/mozilla_b_1_osx.pp
@@ -10,7 +10,7 @@ class roles_profiles::roles::mozilla_b_1_osx {
   include roles_profiles::profiles::network
   include roles_profiles::profiles::ntp
   include roles_profiles::profiles::packages_installed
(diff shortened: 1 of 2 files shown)


aerickson:
We can/should get an official policy on releng on where we need to block public pypyi (ah maybe https://mozilla-hub.atlassian.net/wiki/spaces/ROPS/pages/705527928/Releng+s+Internal+Pypi+Server+-+pypi.pub.build.mozilla.org#What-TC-workers-are-required-to-use-it). Maybe they're going to say VPN needs it also.

I found a way to grab all of the dependencies also... but yeah, it's still a pain.

rcurranmoz:
Closing. New work item https://mozilla-hub.atlassian.net/browse/RELOPS-1475

PR #862, opened 2025-05-15
Decom non-existent pool configs + rm py2 for testers

rcurranmoz:


rcurranmoz:
@aerickson any clues on the test failure? Not sure why it's running `linux_integration_tests-linux-perf-ubuntu-1804`

PR #863, opened 2025-05-16
CHORE: Jira: RELOPS-1264 MSG: Consolidate Win Data and clean up.

markcor:
Roles: All-WIN


markcor:
This includes additional cleanup:
removal of win 11 2009 roles
Consolidation of mozilla_build class into the new single class.

PR #869, opened 2025-05-23
[RELOPS-591] Auto Run Puppet on Macs

rcurranmoz:
Continuation of https://github.com/mozilla-platform-ops/ronin_puppet/pull/824.


modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# Bootstrap a macOS host to a complete Puppet run.
+set -e
+export LANG=en_US.UTF-8
+export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/puppetlabs/bin"
+
+# Puppet configuration
+
+# This will probably need to change in the future to a more generic location
+LOCAL_PUPPET_REPO="/Users/relops/Desktop/puppet/ronin_puppet"
```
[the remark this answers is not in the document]
rcurranmoz: Thank you @aerickson 

I made the suggested changes for your review.

modules/macos_run_puppet/manifests/init.pp
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    file { '/opt/puppet_environments':
+      ensure => directory,
+      owner  => 'root',
+      group  => 'wheel',
+      mode   => '0755',
+    }
+
+    file { '/opt/puppet_environments/ronin_settings':
```
aerickson, line 25: I don't think we want to version/manage this file in puppet, it should just be placed on individual hosts where we want the override. 

(I think you're doing this to avoid hardcoding your test branch... but let's not version it, just place it manually). 

I do think it's a good idea to deliver this file basically as ronin_settings.example... so the format is easy for people to implement when they want to test.

modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,196 @@
+#!/bin/bash
```
aerickson, line 1: Let's reorganize this file and put non-functions after the functions. Everything is intermingled right now and It's a bit tricky to read.

So, this order:

    constant/variable definitions
    functions
    main/non-functions

modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,196 @@
...
+# Ensure Puppet Repository Exists
+get_puppet_repo() {
+    if [ ! -d "$LOCAL_PUPPET_REPO" ]; then
+        fail "Local Puppet repository not found at $LOCAL_PUPPET_REPO"
+    fi
+    cd "$LOCAL_PUPPET_REPO" || fail "Failed to change directory to $LOCAL_PUPPET_REPO"
+
+    # Inject Hiera Secrets
+    mkdir -p ./data/secrets
+    cp /var/root/vault.yaml ./data/secrets/vault.yaml
```
aerickson, line 112: let's do something like `chmod 0600 "${WORKING_DIR}/data/secrets/vault.yaml"` (from the linux run-puppet.sh script) after the copy to just be sure the perms are locked down on these secrets (we shouldn't rely on sane umask values for root being set).

modules/macos_run_puppet/manifests/init.pp
rcurranmoz: New commit pushed resolving this I think. Thank you

modules/macos_run_puppet/files/run-puppet.sh
rcurranmoz: New commit pushed resolving this I think. Thank you

aerickson (commented):
Looking good. Just some tweaks to the refactoring/reordering. 

I'd think about adding email failure notification (see the linux script). It's super handy on the linux hosts to get the digest email with failures every day. Otherwise it's not obvious if there are failures.

modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,209 @@
...
+    # shellcheck source=/opt/puppet_environments/ronin_settings
+    echo "Loading settings from $SETTINGS_FILE..."
+    source "$SETTINGS_FILE"
+else
+    echo "No override settings file found at $SETTINGS_FILE; using script defaults."
+fi
+
+# Change before merge
+: "${PUPPET_REPO:=https://github.com/rcurranmoz/ronin_puppet.git}"
+: "${PUPPET_BRANCH:=local_puppet_work}"
```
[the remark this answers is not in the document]
aerickson: If your testing hosts have an override file, we should be able to go to normal values here now.

modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,209 @@
...
+if [ -f "$SETTINGS_FILE" ]; then
+    # shellcheck source=/opt/puppet_environments/ronin_settings
+    echo "Loading settings from $SETTINGS_FILE..."
+    source "$SETTINGS_FILE"
+else
+    echo "No override settings file found at $SETTINGS_FILE; using script defaults."
+fi
+
+# Change before merge
+: "${PUPPET_REPO:=https://github.com/rcurranmoz/ronin_puppet.git}"
```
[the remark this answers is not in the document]
rcurranmoz: gpt suggestion :) 

image

modules/macos_run_puppet/files/run-puppet.sh
```diff
@@ -0,0 +1,205 @@
...
+    esac
+}
+
+### ---------------------------------------------
+### 3. Main Execution Logic
+### ---------------------------------------------
+
+GIT_USERNAME=$(extract_username_from_url "$GIT_REPO_URL")
+LOCAL_PUPPET_REPO="/opt/puppet_environments/${GIT_USERNAME}/ronin_puppet"
+SETTINGS_FILE="/opt/puppet_environments/ronin_settings"
```
aerickson, line 124: This seems like it could go up top since it's not variable based.

modules/macos_run_puppet/files/ronin_settings.example
```diff
@@ -0,0 +1,10 @@
+# if you place this file at `/opt/puppet_environments/ronin_settings`
+# the `run-puppet.sh` script will use the values here.
+
+# puppet overrides
+PUPPET_REPO='https://github.com/rcurran/ronin_puppet.git'
+PUPPET_BRANCH='local_puppet_work'
+PUPPET_MAIL='rcurran@mozilla.com'
+
+# taskcluster overrides
+# WORKER_TYPE_OVERRIDE='gecko-t-osx-1400-r8-staging'
```
aerickson, line 10: Nice.

Support for WORKER_TYPE_OVERRIDE requires an additional script that I wrote (used in the linux run-puppet.sh) that goes in and does a find/replace in the worker-manager config.

aerickson (commented):
Looking good. It's working well?

aerickson (approved):
Nice! Let's ship it.

I will fix the broken master test in another PR (https://github.com/mozilla-platform-ops/ronin_puppet/pull/870).

PR #921, opened 2025-08-12
mac workers: fix bug in worker-runner.sh

aerickson:
Fix bug introduced in https://github.com/mozilla-platform-ops/ronin_puppet/pull/918.


aerickson:
Tested with an override file on macmini-r8-379.test.releng.mdc1.mozilla.com and it's working.

PR #945, opened 2025-09-18
remove m2 reference

rcurranmoz:


rcurranmoz:
@aerickson can you take a look?

PR #1108, opened 2026-03-16
[RELOPS-2269] Skip people remover script when no listed users exist

rcurranmoz:
Add an unless guard that checks for each user's home directory — exits 0 (skip exec) when none are present, exits 1 (run exec) if any remain. This avoids running the script on every puppet apply once cleanup is done.


aerickson (dismissed):
Would be nice to have the names of users to purge in a variable. The cleanup script must have the names, maybe we should just move the logic into the remover script.

LGTM

rcurranmoz:
Good call @aerickson — moved the skip logic into `user_cleanup.sh --check` so the user list lives only in the script (the `peopleToRemove` array) instead of being duplicated in the puppet `unless`. The exec now just does `unless => "${remover_script} --check"`. PTAL.

rcurranmoz:
tested across macmini-r8-177, macmini-r8-63, and macmini-r8-63

jwmossmoz (approved):

PR #1110, opened 2026-03-16
[RELOPS-2269] Fix remove_old_hg triggering on puppet-managed symlink

rcurranmoz:
The onlyif 'test -f' check matched the /usr/local/bin/hg symlink created by the File resource, causing puppet to delete and recreate it on every apply. Narrow the check to regular files only so the exec only fires when the old hg binary (not a symlink) is present.


aerickson (approved):
nice

rcurranmoz:
sanity checked against macmini-m4-95, macmini-r8-147,  and macmini-r8-165 (r8 + arm)

PR #1145, opened 2026-03-24
[RELOPS-1955] Automate landing new Taskcluster binaries on mac workers

rcurranmoz:
## What this does

Automates landing Taskcluster binaries (`generic-worker`, `start-worker`, `livelog`, `taskcluster-proxy`) on the macOS CI workers, replacing the manual S3-upload workflow.

Today binaries are installed via `packages::macos_package_from_s3`, which requires someone to **manually upload each new Taskcluster release** to the `ronin-puppet-package-repo` S3 bucket before workers can use it. This PR adds `packages::macos_taskcluster_binary`, which pulls binaries straight from the upstream GitHub release and keeps the S3 bucket only as a fallback.

## How it works

`worker_runner` (`modules/worker_runner/manifests/init.pp`) iterates the Taskcluster binary set and calls the new define for each one:

- **Primary source — GitHub release:**
  `https://github.com/taskcluster/taskcluster/releases/download/v<version>/<asset>-darwin-<arch>`
- **Fallback source — S3 (same layout the old define used):**
  `https://s3-us-west-2.amazonaws.com/ronin-puppet-package-repo/macos/public/common/<bin>-<version>-<arch>`
- Architecture is auto-detected (`arm64` for Apple Silicon, `amd64` for Intel).
- Handles the `generic-worker-simple` → `generic-worker-insecure` asset rename introduced in Taskcluster v62.
- Uses `curl -f` so a missing asset exits non-zero (triggering the fallback) instead of silently saving a 404 error page as the "binary".

**Net effect:** to land a new Taskcluster version, just set the worker's `taskcluster_version` and puppet downloads it on the next run — no more manual S3 uploads. The S3 bucket still covers us if GitHub is unavailable (for versions that have been uploaded there).

## Version pins included
- `gecko_t_osx_1500_m4`, `_m4_ipv6`, `_m_vms` → **91.0.2**
- `gecko_t_osx_1500_m4_staging` → **100.4.0**
- `gecko_t_osx_1400_r8_staging` → **97.0.1**

## macOS compatibility note
macOS 10.15 hosts (the `1015` fleet) can only run the **60.3.4** line — Taskcluster builds from ~v62+ are compiled with a macOS 12.0 minimum and abort on 10.15 (`dyld: Symbol not found: _SecTrustCopyCertificateChain`). The macOS 14 (`1400`) and 15 (`1500`/m4) fleets run current versions fine.

## Operational note on version source
On the masterless hosts, `worker.taskcluster_version` is resolved from the host's `/var/root/vault.yaml`, which takes precedence over the role yaml. So the role-yaml versions above apply via Bolt, but on-host the version is whatever `vault.yaml` pins. This PR does **not** change that — it only changes *how* the binaries are fetched, not *which* version is selected.

## Testing
Validated on staging hosts `macmini-r8-126` (10.15 / amd64), `macmini-r8-375` (14 / amd64), and `macmini-m4-111` (15 / arm64):
- Correct-arch binaries download and run across 60.3.4 / 91.0.2 / 95.0.3 / 97.0.1 / 100.4.0
- GitHub → S3 fallback verified (forced a GitHub miss, binary came from S3)
- Clean-failure path verified (both sources miss → no corrupt file written)


aerickson (dismissed):
lgtm

jwmossmoz (approved):

PR #1152, opened 2026-03-26
Enable Safari Remote Automation on SIP-enabled macOS 14/15 workers

rcurranmoz:
## What this changes

- **macos_safaridriver / Safari remote automation:** replace the previous bash-wrapped `osascript` invocation with a LaunchAgent that runs `osascript` directly against an applescript file. Avoids TCC attribution to bash (which never resolves cleanly on macOS 14/15) and keeps GUI session context.
- **New deploy artifacts:** `safari-enable-remote-automation.applescript`, `com.mozilla.safari.enableautomation.plist`, and parallel Safari Technology Preview variants. The applescript writes a semaphore file (content = `1`) on success so puppet can idempotently detect a completed run.
- **macos_safaridriver::init** bootstraps the LaunchAgent into `gui/<uid>` on Darwin 23/24 (macOS 14/15), polls for the semaphore for up to 180s, and falls back to the legacy `launchctl asuser` flow on Darwin 19–22 (macOS 10.15–13). Safari TP block is `onlyif` Safari TP is installed.
- **macos_tcc_perms / macos_safaridriver/add_tcc_perms:** make the macOS 14/15 system-DB writes **SIP-aware** at runtime (`csrutil status | grep -qi "disabled"`). SIP-disabled hosts continue writing the system TCC.db directly — preserves the legacy fleet's behavior. SIP-enabled hosts skip the system-DB writes (those would silently fail on macOS 14/15 anyway) and rely on the new MDM PPPC profile for system-level grants. User-DB writes happen on every host.
- **New `org.mozilla.ci-tcc-pppc.mobileconfig`:** uploaded to SimpleMDM as a Custom Configuration Profile to supply the system-level TCC grants that SIP-enabled hosts can no longer get from sqlite3 writes.
- **Idempotency semaphores:** both `tcc_perms.sh` and `add_tcc_perms.sh` now write `/var/tmp/semaphore/tcc-perms-applied` and `/var/tmp/semaphore/safari-tcc-perms-applied` respectively on success. The corresponding puppet exec resources gain an `unless` clause so the scripts run once per host until the operator clears the semaphore.
- **macos_run_puppet/run-puppet.sh:** TCC.db transient-inaccessibility no longer triggers a reboot — only a missing file does. Reduces noisy reboots when puppet briefly can't open cltbld's TCC.db.
- **cltbld_user:** skip the `dscl AuthenticationAuthority` exec on macOS 14/15 — the key doesn't exist in the directory record on those versions and `dscl -create` returns `eDSPermissionError`. Authentication works without it.

## Validation

Tested against 11 live hosts spanning all three macOS versions in the fleet by setting `PUPPET_BRANCH=sip-compatible-safari-automation` and running `/usr/local/bin/run-puppet.sh`, then verifying post-state on disk.

### Hosts tested

| Pool | Hosts | OS | SIP state |
|---|---|---|---|
| `gecko-t-osx-1500-m4-no-sip` | `macmini-m4-{111,112,113}` | 15.3 | disabled |
| `gecko-t-osx-1400-r8-staging` | `macmini-r8-{246,361,375}` | 14.7.5 | disabled |
| `gecko-t-osx-1400-r8-staging` | `macmini-r8-{109,154}` | 14.7.5 | "unknown (Custom Configuration)" |
| `gecko-t-osx-1015-r8-staging` | `macmini-r8-{125,126,127}` | 10.15.7 | disabled |

### Per-host verification

For each host: `puppet apply` succeeded with no errors. Post-state verified via:

- `/var/tmp/semaphore/tcc-perms-applied` written
- `/var/tmp/semaphore/safari-tcc-perms-applied` written (macOS 14/15 only — gated to that branch in `add_tcc_perms.sh`)
- `/Users/cltbld/Library/Preferences/semaphore/safari-enable-remote-automation-has-run` written with content `1` (= applescript success marker)
- TCC system-DB row counts unchanged on SIP-disabled hosts (legacy writes preserved)
- TCC user-DB has new `/usr/bin/osascript` Accessibility/AppleEvents/PostEvent grants on every host
- On macOS 10.15 hosts: confirmed the legacy `launchctl asuser` Safari flow ran (no new LaunchAgent plists deployed in cltbld's home dir, legacy `/var/tmp/semaphore/safaridriver-enabled` preserved)

### Reproducing the validation

On any single host:

```bash
# Pin the host to this PR's branch and trigger a puppet apply.
sudo bash -c "PUPPET_BRANCH=sip-compatible-safari-automation /usr/local/bin/run-puppet.sh"

# Verify with sudo (cltbld's home is not readable to other users):
sudo test -e /var/tmp/semaphore/tcc-perms-applied                                                  && echo OK
sudo test -e /var/tmp/semaphore/safari-tcc-perms-applied                                          && echo OK   # macOS 14/15 only
sudo cat   /Users/cltbld/Library/Preferences/semaphore/safari-enable-remote-automation-has-run    # should print "1" on macOS 14/15
```

## Known limit

The SIP detection regex (`csrutil status | grep -qi "disabled"`) does not match the literal output `"unknown (Custom Configuration)"` that some hosts return when SIP is in a partially-customized state (observed on `r8-109` and `r8-154` during validation). Those hosts get treated as SIP-enabled and the new SIP-aware code path skips the system-DB writes they currently rely on. Existing system-DB entries from prior master-branch puppet runs persist, so functionality is not actively broken today — but the entries won't be re-applied if anything ever clears TCC.db on those hosts.

Trivial follow-up fix that closes the gap:

```bash
# Default to legacy (SIP-off) behavior; only treat as SIP-on when csrutil
# clearly says "enabled". Catches "Custom Configuration" and other edge cases.
SIP_DISABLED=true
if csrutil status 2>/dev/null | grep -qiE "status:[[:space:]]*enabled\.?$"; then
    SIP_DISABLED=false
fi
```

## Operational notes for after merge

- `tcc_perms.sh` and `add_tcc_perms.sh` are now one-shot per host (gated by their respective `/var/tmp/semaphore/` files). If TCC.db ever needs to be re-applied (e.g. after EACS, manual reset, or unexpected corruption), the operator must `rm` the semaphore to re-trigger the script on the next puppet run.
- The `org.mozilla.ci-tcc-pppc.mobileconfig` profile needs to be uploaded to SimpleMDM and assigned to the SIP-enabled host group before SIP-enabled hosts will get the system-level grants. The PR adds the file; SimpleMDM assignment is a separate operator step.


aerickson (dismissed):
nice

aerickson (dismissed):
LGTM, just one note about inventory.

rcurranmoz:
target host: macmini-m4-118.test.releng.mdc1.mozilla.com

modules/macos_safaridriver/HANDOFF.md
```diff
@@ -0,0 +1,143 @@
...
+   Developer in cltbld's VNC session.
+5. If anything is weird, `/var/log/m4-bootstrap.log` (one-shot setup)
+   and `/var/log/m4-bootstrap-driver.log` (boot-by-boot puppet) are
+   your friends 🪵. SSH keys and LaunchDaemon plists self-clean on
+   success; if they're still on disk, the driver hasn't declared
+   victory yet.
+
+---
+
+Good luck. May your AppleScript clicks actually persist. 🍀
```
aerickson: Do you want this file included/committed?

provisioners/macos/simplemdm_bootstrap_sip_safari.sh
```diff
@@ -0,0 +1,250 @@
+#!/bin/bash
```
aerickson, line 1: Is this just a testing script? Maybe move into subdir or note in comment?

aerickson (dismissed):
Previous comments are non-blocking nits. LGTM.

modules/macos_safaridriver/HANDOFF.md
rcurranmoz: Good call — removed in 7052955. It was session-handoff notes that don't belong in the puppet codebase.

provisioners/macos/simplemdm_bootstrap_sip_safari.sh
rcurranmoz: Good catch — moved to provisioners/macos/ in 7052955, alongside the existing bootstrap_catalina.sh and bootstrap_mojave.sh. Renamed to use underscores for naming consistency, and updated the header to explicitly note it's the SimpleMDM script-job source-of-truth and not a puppet-deployed file.

aerickson (approved):

PR #1178, opened 2026-04-22
Grant ScreenCapture and Microphone TCC permissions to generic-worker-multiuser

rcurranmoz:
## Summary

- `generic-worker-multiuser` had `kTCCServiceScreenCapture` explicitly **denied** (`auth_value=0`) in the system TCC database on macOS 15 M4 workers, causing Firefox CI tests using `getDisplayMedia()` and `getUserMedia()` to fail (Bug 2019625)
- `kTCCServiceMicrophone` was similarly denied in the cltbld user TCC database
- Adds both to the macOS 14/15 SQL queries in `tcc_perms.sh`, matching the pattern already used for `start-worker`

MDM PPPC profiles cannot force-grant `ScreenCapture` on macOS 15 (Apple restricts it to `AllowStandardUserToSetSystemService` only), so direct TCC database writes are the only viable path.

**Note:** The `CodeRequirement` uses the cdhash of the current ad-hoc signed binary on the M4 fleet (`a7add5673f62ad48b6acc24c066a9bc966b68acc`). If `generic-worker-multiuser` is ever rebuilt and redeployed, this entry will need updating. A proper long-term fix would be to sign the binary with an Apple Developer certificate so a Team ID-based requirement can be used instead.

Tested on `macmini-m4-114.test.releng.mdc1.mozilla.com` — both permissions confirmed at `auth_value=2` after running the script.

## Test plan

- [ ] Run `tcc_perms.sh` via Puppet on an M4 staging worker and verify `auth_value=2` for both `kTCCServiceScreenCapture` (system TCC) and `kTCCServiceMicrophone` (cltbld user TCC)
- [ ] Run a Firefox CI task on the worker that exercises `getDisplayMedia()` and confirm it no longer fails with a permission error


aerickson (approved):
lgtm

PR #1125, opened 2026-03-18
[RELOPS-2274] Manage git on macOS ci workers

rcurranmoz:

PR #1183, opened 2026-04-23
fix build_and_install_git: remove cwd, cd in command instead

rcurranmoz:
## Summary

- Puppet validates the `cwd` attribute existence **before** running the `unless` check on an exec resource
- This caused `/Stage[main]/Packages::Git/Exec[build_and_install_git]` to fail with `Working directory /tmp/git-2.47.1 does not exist!` even on hosts where git was already installed (the `unless` would have skipped the exec, but Puppet errors out first)
- Fix: remove `cwd => $src_dir` and embed `cd ${src_dir} &&` at the start of the command

## Test plan

- [ ] Verify Puppet apply succeeds on a host where git 2.47.1 is already installed (exec skipped cleanly)
- [ ] Verify Puppet apply builds and installs git on a fresh host without the working directory error


rcurranmoz:
Closing - the original branch was already merged to master as #1125. Opening a new hotfix PR directly against master instead.

PR #1187, opened 2026-04-27
Add cltbld_user profile to gecko-{1,3}-b-osx-arm64-vms roles

rcurranmoz:
## Summary
- `multiuser-static` generic-worker requires the `cltbld` user and `~/Library/LaunchAgents/` to exist before `worker_runner` can write the genericworker LaunchAgent plist
- Both `gecko_1_b_osx_arm64_vms` and `gecko_3_b_osx_arm64_vms` roles were missing `include roles_profiles::profiles::cltbld_user`, causing Puppet to fail with "No such file or directory" on the LaunchAgent plist during VM image builds
- Fix mirrors what `gecko_t_osx_1500_m_vms` already does

## Test plan
- [ ] Builder VM image build (`Build macOS Builder ARM64 VMs` workflow in `macos-vms`) passes Puppet phase 1 with this branch
- [ ] gecko-1b and gecko-3b VMs boot and register with Taskcluster
- [ ] Try push confirms jobs run successfully

### DIFF
--- a/modules/roles_profiles/manifests/roles/gecko_1_b_osx_arm64_vms.pp
+++ b/modules/roles_profiles/manifests/roles/gecko_1_b_osx_arm64_vms.pp
@@ -4,6 +4,7 @@
 
 class roles_profiles::roles::gecko_1_b_osx_arm64_vms {
   include macos_utils::disable_bluetooth_setup
+  include roles_profiles::profiles::cltbld_user
   include roles_profiles::profiles::macos_disable_firewall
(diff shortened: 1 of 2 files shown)


rcurranmoz:
Closing — the root cause was the vault file having `generic_worker_engine: multiuser-static` in the `worker` section, which overrode the Hiera role data via dot-notation traversal. Fixed by removing the `worker` section from the vault files on the runner. Builders use the `multiuser` engine (new user per task), so `cltbld_user` profile is not needed.

PR #1209, opened 2026-05-12
safaridriver: only write has-run semaphore on actual success

### LINKED ISSUE #1207: macos_safaridriver: only write has-run semaphore on actual success
## Bug

The \`safari-enable-remote-automation-has-run\` and \`safari-tech-preview-enable-remote-automation-has-run\` semaphore files at \`/Users/cltbld/Library/Preferences/semaphore/\` are written regardless of whether the underlying \`safari-enable-remote-automation.sh\` script actually succeeded.

The failure case is common on first bootstrap: cltbld's TCC.db doesn't exist yet (cltbld hasn't auto-logged in by the time puppet first tries to grant TCC perms), the safari script errors out, but the semaphore still gets written. Subsequent puppet runs see the semaphore and skip the script, so the worker permanently lacks the safari automation entitlement — silently broken.

Current workaround: hand-remove both files on every newly-provisioned M4 worker so the next reboot's puppet run actually re-applies the fix. Did this on all 20 hosts during the 2026-05-12 m4-130..149 rollout.

## Proposal

The semaphore should be written by the script itself **only** after \`safaridriver --enable\` (or whatever the underlying success indicator is) returns 0 — not by puppet's \`Exec\` resource as a side effect of running.

## Acceptance criteria

- [ ] \`safari-enable-remote-automation.sh\` writes the semaphore at the very end, only on \`set -e\` clean exit
- [ ] Same for the tech-preview variant
- [ ] Puppet's \`creates =>\` (or equivalent unless guard) still works
- [ ] Verified that a failed run leaves no semaphore, so the next puppet apply will retry

## Related

[[Bug 2019466 / D295834 / D295832 / PR #1178]] — same general TCC-perms-on-first-boot family.

rcurranmoz:
## Summary

Fixes #1207.

The script `modules/macos_safaridriver/files/safari-enable-remote-automation.sh` currently writes the `safari-(tech-preview-)enable-remote-automation-has-run` semaphore unconditionally, even when the underlying osascript "click the checkbox" sequence failed (most commonly because cltbld's TCC.db doesn't exist yet on first bootstrap). Puppet's `unless => /bin/test -f <semaphore>` then sees the file on the next apply and skips the resource — the worker is permanently broken with no retry path.

This PR:

- Removes the pre-emptive `touch` of the semaphore file at script startup
- Refactors `enable_remote_automation` to return non-zero on failure
- Only writes the semaphore on confirmed osascript success
- Aggregates per-app return codes so the script exits non-zero if Safari and/or Safari Tech Preview failed (puppet will surface this as a visible error)

With this change, puppet's existing `unless` guard does the right thing: the semaphore is created only when configuration actually succeeded, so a failed run is retried on the next puppet apply.

## Test plan

- [ ] Manually run on a fresh worker where cltbld TCC.db hasn't been created yet — confirm semaphore is NOT written, script exits non-zero
- [ ] Manually run on a fresh worker after cltbld has auto-logged in — confirm semaphore IS written, script exits 0
- [ ] Re-run puppet after a successful semaphore write — confirm the exec is skipped via `unless`
- [ ] On a 15.x host with both Safari and STP installed, simulate a partial failure (e.g. quit one of the apps mid-osascript) and confirm only the failing app's semaphore is missing

### DIFF
--- a/modules/macos_safaridriver/files/safari-enable-remote-automation.sh
+++ b/modules/macos_safaridriver/files/safari-enable-remote-automation.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -uo pipefail
 
 # Ensure the logged-in user is 'cltbld'
 current_user=$(id -u -n)
@@ -7,7 +8,12 @@ if [ "$current_user" != "cltbld" ]; then
   exit 1
 fi
 
-# Function to enable remote automation for Safari
+# Track per-app failures so puppet's `unless => test -f <semaphore>` can keep retrying
+# on subsequent runs until the work actually succeeds.
+overall_rc=0
+
+# Function to enable remote automation for Safari.
+# Returns 0 on success (semaphore written), non-zero on failure (semaphore NOT written).
 enable_remote_automation() {
   local semaphore_file="$1"
   local semaphore_version="1"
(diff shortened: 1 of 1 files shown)


rcurranmoz:
Superseded by #1152. The verify-after-click gate in #1152's Safari applescripts only writes the success semaphore when the post-click checkbox reads 1, which solves this same silent-failure problem differently. Closing without merging.

Issue #1206, opened 2026-05-12
macos_run_puppet: make bootstrap reboot-survivable via LaunchDaemon


rcurranmoz:
## Motivation

Today bootstrapping a fresh M4 worker requires an external SSH session that babysits the run:
1. SSH in, run \`run-puppet.sh\`
2. First puppet apply fails on cltbld TCC.db → script does \`shutdown -r now\` and exits
3. SSH session dies
4. Wait for host to come back (5–30 min depending on whether MDM also wants to upgrade it)
5. SSH back in, re-run \`run-puppet.sh\`
6. Drop \`/var/tmp/semaphore/run-buildbot\` and reboot

This means every new-host rollout needs a persistent babysitter (a shell loop, a Bolt session, or in today's case a Claude Code orchestration session) for the entire 20–60 minute window. It also can't survive an MDM-triggered OS upgrade mid-bootstrap, which we hit on 12/20 hosts during the macmini-m4-130..149 rollout on 2026-05-12.

## Proposal

Have \`run-puppet.sh\` (or a thin wrapper invoked once by MDM/Bolt) install itself as a LaunchDaemon on first invocation. The LaunchDaemon re-runs puppet on every boot until two conditions are met:
- Puppet apply exits cleanly (no \`Error:\` in log, no TCC.db error)
- \`/var/tmp/semaphore/run-buildbot\` exists

When both are satisfied, the LaunchDaemon unloads and removes itself.

This means a single MDM script-job (or single \`bolt run\` call) can kick the bootstrap and walk away — no SSH session to lose.

## Acceptance criteria

- [ ] On first invocation, \`run-puppet.sh\` writes \`/Library/LaunchDaemons/com.mozilla.ronin-puppet-bootstrap.plist\` referencing itself
- [ ] LaunchDaemon \`RunAtLoad=true\`; runs as root
- [ ] Script is idempotent — surviving the LaunchDaemon trying to start it while a previous instance is still running, e.g. via a pidfile or \`flock\`
- [ ] On successful puppet apply + semaphore present, the script removes the plist and \`launchctl unload\`s it
- [ ] Tested across: TCC.db reboot, MDM-driven OS upgrade reboot, hard power-cycle mid-apply
- [ ] Doesn't fight with the existing periodic puppet run mechanism

## Related

Discovered during M4 fleet rollout (m4-130..149) where 12 hosts hit an MDM 15.3 upgrade mid-bootstrap and needed manual respawn.

[...]

PR #1211, opened 2026-05-12
tcc: gate cltbld-TCC.db-dependent resources on a custom fact

### LINKED ISSUE #1208: macos_run_puppet: replace TCC.db log-grep reboot trigger with cltbld-session gate
## Current behavior

\`run-puppet.sh\` greps puppet's apply output for a specific string:
\`\`\`
unable to open database \"/Users/cltbld/Library/Application Support/com.apple.TCC/TCC.db\"
\`\`\`
…and on match issues \`sudo shutdown -r now\` to force a reboot, on the theory that after reboot cltbld auto-logs in, the TCC.db materializes, and the next puppet apply will succeed.

## Problems

1. **It's a log-grep against a string Apple controls.** If the error message changes across macOS versions (or localizations), we silently lose the reboot trigger.
2. **It reboots in the middle of a puppet apply** rather than at a well-defined boundary, so other resources downstream of the failing TCC resource never get applied in this pass.
3. **No way to distinguish "cltbld TCC.db not ready yet" from "cltbld TCC.db is permanently broken."** Same error string, very different remediation.
4. **The fix-by-reboot is racy**: if cltbld autologin is slow/delayed, the next puppet apply will hit the same error and reboot loop.

## Proposal

Gate the TCC perms resources on cltbld actually having an active console session (or TCC.db existing). Options, increasing in invasiveness:

**A) Conditional skip in puppet:** Use a custom fact (\`cltbld_tcc_db_present\`) that's true when \`/Users/cltbld/Library/Application Support/com.apple.TCC/TCC.db\` exists. TCC perm resources only apply when fact is true. Puppet completes without errors on first pass; second pass (after reboot + autologin) actually applies them.

**B) Wait-loop in the underlying shell scripts** (\`tcc_perms.sh\`, \`add_tcc_perms.sh\`, \`safari-enable-remote-automation.sh\`): poll for TCC.db up to N seconds before failing. Less invasive but ties up the puppet run.

**C) Keep the reboot trigger but base it on the same fact** — detect that cltbld isn't logged in and reboot at the *end* of the apply rather than mid-run.

Option A is the cleanest and meshes with the [[reboot-survivable bootstrap]] work.

## Acceptance criteria

- [ ] First puppet apply on a fresh host completes without a forced reboot from the script
- [ ] TCC-dependent resources cleanly defer to the next apply pass
- [ ] After cltbld autologin, next puppet apply succeeds with no further reboots

## Related

Discovered during M4 fleet rollout 2026-05-12; pairs with [[#1206 reboot-survivable bootstrap]] and the [[safari-automation semaphore bug #1207]].

rcurranmoz:
## Summary

Fixes #1208.

Replaces the log-grep reboot trigger in `run-puppet.sh` with a proactive fact-based gate.

The bootstrap script currently does this:

\`\`\`bash
if grep -q "unable to open database \\"/Users/cltbld/.../TCC.db" "\$TMP_LOG"; then
    echo "Detected TCC.db issue. A reboot is required."
    sudo shutdown -r now
    exit 0
fi
\`\`\`

That string is Apple-controlled (we'd silently lose the trigger if Apple rewords or localizes it), reboots mid-apply (so downstream resources never run in this pass), and can't distinguish "cltbld TCC.db not ready yet" from "cltbld TCC.db permanently broken" — same string, very different remediation.

## Changes

- **New custom fact** \`cltbld_tcc_db_present\` in \`modules/macos_tcc_perms/lib/facter/\`. Boolean — true iff \`/Users/cltbld/Library/Application Support/com.apple.TCC/TCC.db\` exists.
- **\`macos_tcc_perms\`**: gates the \`execute tcc perms script\` exec on \`\$facts['cltbld_tcc_db_present']\`.
- **\`macos_safaridriver\`**: gates \`execute perms script\` and \`execute enable remote automation script\` on the same fact (both write to or depend on cltbld's TCC.db).
- **\`run-puppet.sh\`**: removes the log-grep block. Adds a post-success check — if cltbld TCC.db is still missing after a clean apply, the script reboots once so autologin can fire; the next apply (via bootstrap LaunchDaemon #1206 or the regular at-boot mechanism) will see the fact flip to true and apply the resources cleanly.

## Behavior on a fresh host

1. First apply: autologin set up, TCC resources skipped (no errors)
2. Post-success check sees TCC.db absent → reboot
3. cltbld autologs in → TCC.db materializes
4. Next apply: TCC resources actually run

## Test plan

- [ ] Fresh M4 host (cltbld absent at start of apply): puppet apply completes without errors, post-success check triggers reboot, second apply applies TCC perms
- [ ] Already-bootstrapped host (cltbld TCC.db present): TCC resources apply normally, no reboot
- [ ] Kitchen mac suite: \`running_in_test_kitchen\` guard still keeps TCC resources from firing
- [ ] Confirm \`facter -p cltbld_tcc_db_present\` returns the correct boolean on a managed host
- [ ] Linux suites unaffected (fact is confined to Darwin)

## Related

Pairs naturally with #1206 (reboot-survivable bootstrap LaunchDaemon) — together they eliminate the babysitter ssh pattern entirely.

### DIFF
--- a/modules/macos_run_puppet/files/run-puppet.sh
+++ b/modules/macos_run_puppet/files/run-puppet.sh
@@ -142,12 +142,6 @@ run_puppet() {
     retval=$?
     PUPPET_RUN_DURATION=$SECONDS
 
-    if grep -q "unable to open database \"/Users/cltbld/Library/Application Support/com.apple.TCC/TCC.db" "$TMP_LOG"; then
-        echo "Detected TCC.db issue. A reboot is required."
-        sudo shutdown -r now
-        exit 0
-    fi
-
     if grep -q "^Error:" "$TMP_LOG"; then
         retval=1
     fi
@@ -284,5 +278,16 @@ done
 # Clean Up & Finish
 rm -rf "$TMP_PUPPET_DIR"
 echo "System Installed: $(date)" >> /etc/issue
+
+# If cltbld's user TCC.db doesn't exist yet, autologin hasn't fired. The
+# cltbld_tcc_db_present fact gated off TCC-writing resources in this apply
+# pass, so they haven't run. Reboot so cltbld autologs in and the next
+# puppet apply (via either the bootstrap LaunchDaemon or the regular
+# at-boot LaunchDaemon, see #1206) picks them up.
+if [ ! -f "/Users/cltbld/Library/Application Support/com.apple.TCC/TCC.db" ]; then
+    echo "cltbld TCC.db not present after successful puppet apply; rebooting so autologin can fire and TCC resources can apply on the next pass."
+    sudo shutdown -r now
+    exit 0
+fi
 
 exit 0
--- a/modules/macos_safaridriver/manifests/init.pp
+++ b/modules/macos_safaridriver/manifests/init.pp
@@ -31,8 +31,10 @@ class macos_safaridriver (
             mode    => '0755',
           }
 
-          # TCC.db doesn't exist yet in ci, so skip running the script
-          if $facts['running_in_test_kitchen'] != 'true' {
+          # cltbld's user TCC.db only exists after cltbld first logs in.
+          # Skip these resources on first bootstrap; the next puppet apply
+          # (post-cltbld-autologin reboot) will pick them up.
+          if $facts['running_in_test_kitchen'] != 'true' and $facts['cltbld_tcc_db_present'] {
             exec { 'execute perms script':
               command     => $perm_script,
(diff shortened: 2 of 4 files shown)


rcurranmoz:
Superseded by #1152, whose bootstrap runs puppet in the sshd user-session TCC domain and is reboot-survivable, replacing the cltbld-TCC.db custom-fact gate approach. Closing without merging.

PR #1227, opened 2026-05-29
linux 2404: install openbox and use Intel GPU (bug 2031822)

aerickson:
- Adds a linux_packages::openbox class and installs openbox on Ubuntu 24.04 Talos X11 and Wayland workers. 
- Forces xorg to use the Intel Iris graphics card.

Also includes the linux_perf_profiling profile on the Ubuntu 24.04 Talos Wayland role (missed in earlier patch).

See https://bugzilla.mozilla.org/show_bug.cgi?id=2031822 and https://mozilla-hub.atlassian.net/browse/RELOPS-2363.

### DIFF
--- a/modules/linux_gui/files/20-intel-primary.conf
+++ b/modules/linux_gui/files/20-intel-primary.conf
@@ -1,0 +1,6 @@
+Section "Device"
+    Identifier "Intel P580"
+    Driver "modesetting"
+    BusID "PCI:0:2:0"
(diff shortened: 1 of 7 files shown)


92kns:
likely won't use openbox but i think since the config is already here it would be useful to have that deployed, and it should be harmless to keep.

jwmossmoz (approved):

rcurranmoz (approved):

PR #1231, opened 2026-06-04
Add tart_worker role for gecko-t-osx-1500-m-vms tester VM hosts

rcurranmoz:
## What

Adds the `tart_worker` role + hiera for the macOS Tart **tester** VM hosts
(pool `gecko-t-osx-1500-m-vms`), and reworks the shared `profiles::tart` so
those hosts can be managed on **macOS 15** without the failures we hit during
the m4-235–239 rollout.

## Why

The tester VM hosts ran `tart_worker` off a **private fork branch**
(`rcurranmoz/ronin_puppet@tart_worker_prod`) the hosts can't even re-fetch.
This brings the role onto the canonical repo. While provisioning 5 new hosts on
macOS 15.3 we found two OS-version landmines that the older-built 185–192 never
hit, so the profile needed to flex:

1. **`tart pull` only works from the console GUI session on macOS 15** — Local
   Network privacy denies Tart's connection to the LAN OCI registry from a
   headless puppet/ssh context (`su - admin`), surfacing as the misleading
   `Error: The Internet connection appears to be offline.` (curl from the same
   host pulls fine). So puppet **cannot** own the first pull on 15.x.
2. **gui-domain LaunchAgents can't be loaded over ssh on 15.3** (`bootstrap`
   EIO), so the VMs never start headlessly and don't survive reboot.

## Changes

- **`profiles::tart`** — two new hiera knobs, both defaulting to the current
  gecko-1b builder behaviour (builder hiera untouched → no behaviour change):
  - `tart.manage_image` (default `true`): when `false`, skip pull/clone. Image
    is seeded once by hand from the console; puppet only manages tart + the
    launchd unit. Also stops puppet retry-looping the doomed pull (that caused
    a ~few-hundred-email alert storm from one host over the weekend).
  - `tart.launchd_type` (default `'agent'`): `'daemon'` writes a **system
    LaunchDaemon** (UserName=admin, RunAtLoad, KeepAlive) loaded via
    `launchctl bootstrap system` — starts at boot headlessly, survives reboot.
  - Fixes a latent bug in `pull_initial_image`'s `unless` (awk'd `$1` = literal
    "local" instead of the VM-name field, so it never matched and re-pulled
    every apply).
- **New template** `com.mozilla.tartworker.daemon.plist.epp` (the LaunchDaemon).
- **`data/roles/tart_worker.yaml`** sets `manage_image: false` + `launchd_type: daemon`
  + tester registry/image (`10.49.56.161` / `sequoia-tester`).

## Validation / rollout notes

- `puppet parser validate`, `puppet epp validate`, pre-commit: all pass.
- **Still needs a real-host validation run**: confirm the system LaunchDaemon
  actually boots `tart run` as admin and the worker registers (builder pool
  precedent says daemon-run works, but it should be verified on one tester host).
- Rollout: seed the image once from the console (`vms.sh`), then `run-puppet.sh`
  applies the role → installs tart + writes the LaunchDaemon + bootstraps it.
- Existing m4-235–239 were brought up manually (`tart run &`); after this lands
  + a puppet apply they become reboot-persistent. m4-238/239 have root-owned
  VMs (provisioned via sudo) and must be redone admin-owned for the
  UserName=admin daemon to find them.

### DIFF
--- a/data/roles/tart_worker.yaml
+++ b/data/roles/tart_worker.yaml
@@ -1,0 +1,23 @@
+---
+ntp_server: time.apple.org
+
+# Puppet agent
+puppet_agent:
+  package_version: 7.28.0
+
+puppet_run_strategy: cron
+
+# Tart VM host configuration: runs gecko-t-osx-1500-m-vms tester VMs.
+# Registry/image differ from the gecko-1b builder hosts (which use 10.49.56.83
+# / sequoia-gecko1b-vms). The worker identity is baked into the sequoia-tester
+# image, so no host-side vault secrets are required.
+tart:
+  version: '2.30.0'
+  registry_host: '10.49.56.161'
+  registry_port: 5000
+  oci_image: 'sequoia-tester'
+  oci_tag: 'prod-latest'
+  vm_name_prefix: 'sequoia-tester'
+  worker_count: 2
+  insecure: true
+  user: 'admin'
--- a/data/roles/tart_worker.yaml
+++ b/data/roles/tart_worker.yaml
@@ -11,6 +11,15 @@ puppet_run_strategy: cron
 # Registry/image differ from the gecko-1b builder hosts (which use 10.49.56.83
 # / sequoia-gecko1b-vms). The worker identity is baked into the sequoia-tester
 # image, so no host-side vault secrets are required.
+#
+# manage_image: false  -> puppet does NOT pull/clone. On macOS 15 `tart pull`
+#   only works from the console GUI session (Local Network privacy), so the
+#   image is seeded once by hand (see tart-provision.sh / the on-host vms.sh)
+#   and puppet only manages tart + the launchd unit. This also stops puppet
+#   from retry-looping the doomed headless pull (that caused an email storm).
+# launchd_type: daemon -> system LaunchDaemon (starts at boot headlessly,
+#   survives reboot) instead of a gui-domain LaunchAgent that can't load over
+#   ssh on 15.x.
 tart:
   version: '2.30.0'
   registry_host: '10.49.56.161'
@@ -21,3 +30,5 @@ tart:
   worker_count: 2
   insecure: true
   user: 'admin'
+  manage_image: false
+  launchd_type: 'daemon'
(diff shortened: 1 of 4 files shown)


rcurranmoz:
## Validated end-to-end on macmini-m4-235 (2026-06-10)

Migrated a real tester host (admin-owned VMs, previously running the gui-domain LaunchAgent) from agent → daemon via a one-shot `puppet apply` of this branch, then rebooted to confirm persistence. Everything below was done over plain SSH — **no VNC/console login** — so anything that comes up is purely the boot-time system LaunchDaemon.

**Migration apply (clean, no errors):**
```
Exec[evict_agent_tartworker_1]            executed successfully   # old gui agent booted out
File[.../LaunchAgents/...-1.plist]         ensure: removed         # can't reload at next boot
File[/Library/LaunchDaemons/...-1.plist]   ensure: defined content
Exec[load_tartworker_1]                    Triggered 'refresh' from 2 events   # daemon bootstrapped
   ... identical for worker 2 ...
Applied catalog in 0.20 seconds
```

**After `sudo reboot` (host back in ~27s), with no console login:**
```
system domain:  308 com.mozilla.tartworker-2   (status 0)
                309 com.mozilla.tartworker-1   (status 0)
tart run procs: PID 308/309, PPID 1 (launchd), user admin   # owned by the daemon, not a gui agent
tart list:      sequoia-tester-1, sequoia-tester-2  → both running
```

**Idempotency:** a second `puppet apply` reported exit 0 (no changes), no `evict`/`load`/error lines, and the daemon PIDs were unchanged (not flapped).

### Two fixes added on top of the original branch after testing exposed them
1. **Evict the old LaunchAgent when switching to `daemon`.** The original daemon path wrote the LaunchDaemon but left the gui-domain LaunchAgent loaded. Every existing tester host (235–239) keeps that agent alive with `KeepAlive`, so it races the daemon for the same VM → `tart run` loses on "VM already running" and `KeepAlive`-flaps. Now `launchctl bootout gui/$(id -u <user>)/<label>` (root can target the user gui domain, `onlyif` keeps it idempotent) + `ensure => absent` on the agent plist, sequenced before the daemon load.
2. **Wrap the bootstrap exec in `/bin/bash -c`.** It used bare shell operators (`;`, `||`, `2>/dev/null`), which Puppet `exec` does not pass through a shell, so it would have been mis-parsed on first apply. Now matches the `install_tart` convention.

The `agent` path (gecko-1b builder hosts) is unchanged.

### Rollout note
236 and 237 (clean, admin-owned) will migrate on a normal puppet run. **238/239 have root-owned VMs** (`/var/root/.tart` from a `sudo` during provisioning), so the admin daemon won't find them — their VMs need to be recreated as admin first.

jwmossmoz (approved):

PR #1252, opened 2026-06-25
macos_screenshot_helper: fix empty/blank failure screenshots via WatchPaths

rcurranmoz:
## Problem

Failure screenshots on the `gecko-t-osx-1500-m4` pool are frequently empty (zero-byte) or show a **blank desktop** instead of the browser. RELOPS-1438's trigger/LaunchAgent mechanism is correctly wired (tasks run as `cltbld`, `cltbld` owns `/dev/console`, trigger paths match), so this is not the original empty-on-ARM TCC problem — it's a **timing race**.

### Root cause (observed on `macmini-m4-102`)
`com.mozilla.screencapture.plist` used `StartInterval=1`, but `capture-on-demand.sh` exits almost instantly when the trigger file is empty. launchd therefore throttles respawns to its ~10s minimum-runtime floor:

```
launchctl print gui/555/com.mozilla.screencapture
    runs = 94          # over 16 min uptime ≈ one run / ~10s, NOT 1s
    minimum runtime = 10
    last exit code = 0
```

`/tmp/screenshot-capture.log` was absent despite 94 runs — every run found the trigger empty and bailed before capturing. The mochitest/reftest harness only waits **5s** (`10 × 0.5s`) for the screenshot, so the agent often doesn't read the trigger before the harness gives up → zero-byte upload; when it fires ~10s late onto an idle screen → blank-desktop capture. (Sampling recent failed tasks: ~58% real content, ~38% blank, plus a band of 0-byte files.)

## Fix

- **`WatchPaths`** on the trigger file instead of `StartInterval` — launchd fires on write with sub-second latency and is not subject to the respawn throttle.
- `capture-on-demand.sh`: capture to a temp file then **atomic `mv`** so the harness never observes a partial/zero-byte frame; clear the trigger up front to prevent double capture.
- Ensure the trigger file exists so the watch is armed before the first task.
- Reload via **bootout + bootstrap** (`kickstart -k` restarts the loaded job but does not re-read an edited plist, so already-provisioned hosts would otherwise keep the old config).

## Testing

- `pre-commit` (puppet-lint, puppet-validate, shellcheck, templated-script checker) passes; `plutil -lint` OK on the plist.
- The race only manifests on real hardware, so this needs validation on a worker: apply, then confirm `WatchPaths` fires (`/tmp/screenshot-capture.log` grows on task failures) and that recent `mozilla-test-fail-screenshot_*.png` artifacts show the browser rather than a blank desktop.

### DIFF
--- a/modules/macos_screenshot_helper/files/capture-on-demand.sh
+++ b/modules/macos_screenshot_helper/files/capture-on-demand.sh
@@ -6,14 +6,24 @@ TRIGGER_FILE="/Users/cltbld/.trigger_screenshot"
 
 OUTPUT_PATH=$(cat "${TRIGGER_FILE}")
 
+# Clear the trigger file up front. The truncation itself fires a WatchPaths
+# event, but that re-run sees an empty trigger and exits, so a single request
+# can never be captured twice.
+: > "${TRIGGER_FILE}"
+
 # Ensure output directory exists
 mkdir -p "$(dirname "${OUTPUT_PATH}")"
 
-# Take screenshot
-/usr/sbin/screencapture -C -x -t png "${OUTPUT_PATH}"
+# Capture to a temp file, then atomically move it into place. The harness
+# polls for OUTPUT_PATH to become non-empty, so it must only ever appear once
+# fully written -- never as a partial or zero-byte frame.
+TMP_PATH="${OUTPUT_PATH}.tmp.$$"
+/usr/sbin/screencapture -C -x -t png "${TMP_PATH}"
 
-# Log the capture
-echo "$(date): Screenshot saved to ${OUTPUT_PATH}" >> /tmp/screenshot-capture.log
-
-# Clear the trigger file
-: > "${TRIGGER_FILE}"
+if [ -s "${TMP_PATH}" ]; then
+  mv -f "${TMP_PATH}" "${OUTPUT_PATH}"
+  echo "$(date): Screenshot saved to ${OUTPUT_PATH}" >> /tmp/screenshot-capture.log
+else
+  rm -f "${TMP_PATH}"
+  echo "$(date): screencapture produced no output for ${OUTPUT_PATH}" >> /tmp/screenshot-capture.log
+fi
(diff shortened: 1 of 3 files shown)


jwmossmoz (approved):

rcurranmoz:
## On-hardware validation ✅

Validated by pinning two staging workers (`macmini-m4-112`, `macmini-m4-113`, role `gecko_t_osx_1500_m4_staging`, aarch64) to this branch via `ronin_settings` and running `run-puppet.sh` **4×** on each.

### Convergence + idempotency (identical on both hosts)

| Run | Exit | screenshot-helper resources changed | apply time |
|----:|:----:|:--|:--|
| 1 | 0 | `File[capture-on-demand.sh]` + `File[com.mozilla.screencapture.plist]` content changed; `Exec[load or restart screenshot agent]` triggered (refresh) | ~11 s |
| 2 | 0 | none | ~2.5 s |
| 3 | 0 | none | ~2.5 s |
| 4 | 0 | none | ~2.5 s |

Changes apply on run 1 only; runs 2–4 make **zero** screenshot-helper changes → clean idempotency. The bootout+bootstrap reload picked up the new plist on these already-provisioned hosts.

### Runtime behavior is the intended fix

**Before** (baseline, `StartInterval=1`):
```
run interval = 1 seconds
runs = 1699          # over 4h44m uptime ≈ one run / ~10s (launchd respawn throttle)
```

**After** (this branch, `WatchPaths`):
```
com.apple.launchd.WatchPaths => { "WatchPaths" => [ /Users/cltbld/.trigger_screenshot ] }
watching = 1
state = not running   # idle until the harness writes the trigger
runs = 1
```

No more `StartInterval`; the agent sits idle and fires on write instead of respawning every ~10 s. Trigger file present/armed (`/Users/cltbld/.trigger_screenshot`, owned `cltbld`); `capture-on-demand.sh` carries the atomic-write logic.

### Not yet observed
An actual task-failure screenshot on the soak workers rendering the browser rather than a blank desktop — the timing race is eliminated by construction, but that end-to-end confirmation requires a real test failure on a pinned worker. Leaving 112/113 pinned for a short soak.

PR #1256, opened 2026-06-26
linux: Recover interrupted dpkg and fetch TC binaries from GitHub

aerickson:
## Summary

  - Recover interrupted dpkg state before running Puppet on Ubuntu hosts.
    - Detect both `dpkg --audit` output and pending files in `/var/lib/dpkg/updates`.
    - Run `dpkg --configure -a` before the next converge so `run-puppet` does not loop forever on apt failures.
    - Add local Kitchen/Ruby setup docs and document how to manually test interrupted dpkg recovery.
  - Add support for fetching Taskcluster binaries from GitHub releases with SHA256 verification.
    - Only used by 24.04 for now. See `taskcluster_binary_source` arg.
  - Skips Chrome where no arm64 package is available (test-kitchen).

  ## Verification

  - Ran pre-commit hooks successfully during commits.
  - Verified `./bin/kitchen_docker converge linux-perf-ubuntu-2404` on local arm64 Docker after the Kitchen support changes.
  - Manually verified interrupted dpkg recovery in Kitchen:
    - Created a real recoverable dpkg state with `dpkg --unpack` on `zstd`.
    - Confirmed `dpkg --audit` reported `zstd` unpacked but not configured.
    - Ran `run-puppet --no-sentinel`.
    - Confirmed `run-puppet` detected the interrupted dpkg state and ran `dpkg --configure -a`.
    - Confirmed `dpkg --audit` was clean afterward and `zstd` returned to `Status: install ok installed`.

  ## Notes

  A fake file in `/var/lib/dpkg/updates` is useful only to prove that apt can emit the interrupted-dpkg error. It is not a good recovery test because `dpkg --configure -a` correctly rejects invalid update files. The README now documents the safer `dpkg --unpack` based  Kitchen test.


aerickson:
- Tested on ms097 (tested dpkg recovery, tc installation still works - not using new method on 1804).
- Tested on ms028 (tested tc installation - uses new method on 2404).

rcurranmoz (approved):
hell yeah

PR #1282, opened 2026-07-15
RELOPS-2467: dynamic xperf trace mechanism in win_hw_profiling

markcor:
## Summary

Implements the spike RELOPS-2467: a **new, dynamic** xperf trace mechanism so developers can choose xperf options **at run time** instead of editing a hardcoded script and re-running Puppet.

Adds `win_hw_profiling::xperf_dynamic_trace` as a fully independent sibling of the existing `win_hw_profiling::xperf_kernel_trace`. **The existing kernel-trace scripts and tasks are left untouched.**

## How it works

- **Puppet only stages the pieces.** It drops `xperf_dyn_{start,stop,register_tasks}.ps1` and registers two SYSTEM scheduled tasks — `xperf_dyn_trace_start` / `xperf_dyn_trace_stop` — with a `BUILTIN\Users` GR/GX ACE so unprivileged task users can *trigger* (not modify) them. Puppet never runs the trace, and tasks re-register only when a script changes — never when options change.
- **The task drives it at run time.** An unprivileged task user writes `C:\Users\<interactive-user>\xperf\xperf_dyn_options.json` (their own profile, which is also the trace output dir), triggers the start task, runs the workload, triggers the stop task, and collects `combined.etl` from the same dir.
- **Trace scope is intentionally unrestricted** — developers may request any providers/kernel flags they need.

## Security / containment

The options file is written by an unprivileged user and consumed by a **SYSTEM** task, so the scripts enforce process containment (the primary review focus):

- The **only** executable ever launched is script-resolved `xperf.exe`; the options file cannot name a program or path to one.
- **The script owns the flags; the file owns only the values.** Validated values fill the `-start/-on/-stackwalk/-f/-BufferSize` slots — the file cannot introduce a new flag or extra argv token.
- xperf is invoked as `& $xperf @args` (argument array, **no shell**) — metacharacters in a value are inert.
- Every value is validated (charset / integer bounds / bare `*.etl` filename, no path traversal). Any failure falls back to built-in defaults matching `xperf_kernel_trace`; the scripts never fail open.

## Changes

- `modules/win_hw_profiling/manifests/xperf_dynamic_trace.pp` — new manifest (stages scripts + refreshonly register exec)
- `modules/win_hw_profiling/files/xperf_dyn_start.ps1` — reads/validates options, builds args, runs xperf
- `modules/win_hw_profiling/files/xperf_dyn_stop.ps1` — stops sessions from the same file, merges `.etl`
- `modules/win_hw_profiling/files/xperf_dyn_register_tasks.ps1` — registers the two SYSTEM tasks
- `modules/win_hw_profiling/README.md` — options schema + security model
- `modules/roles_profiles/manifests/profiles/microsoft_tools.pp` — include the new class on testers (additive)

## Testing / follow-up

- Prototype only; not yet exercised on a live Windows worker. Suggested follow-up: extend the Windows serverspec suite to assert the two tasks register, a valid options file produces the expected trace, and hostile input (path traversal, injected flags, out-of-range buffersize) is rejected → defaults.


markcor:
Related pushes: 

https://treeherder.mozilla.org/jobs?repo=try&revision=ced2930e71c8e7f19bdc937ab0ce51241d7fbc19

https://treeherder.mozilla.org/jobs?repo=try&author=mcornmesser%40mozilla.com&selectedTaskRun=c4eLgQwjSxqSj8xxZ32Pyw.0

aerickson (approved):
nice

PR #1294, opened 2026-07-17
tart: rolling, drain-aware VM image update

rcurranmoz:
## Why

`tart-update-vms.sh` (the "roll a new image onto a host" script) had three problems that made it unsafe/ineffective for the tester VM hosts:

1. **Wrong launchd path** — it used `launchctl load/unload` on `~/Library/LaunchAgents`, but the tester hosts run `launchd_type: daemon` (`bootstrap/bootout system` on `/Library/LaunchDaemons`). So it wouldn't actually cycle those workers.
2. **All-at-once** — stopped *both* VMs, so the whole host went offline during the update.
3. **No drain** — killed VMs regardless of whether they were mid-task, interrupting running Firefox jobs.

## What

Rewrite it as a safe **per-host primitive**:
- **daemon- and agent-aware** (`bootout/bootstrap system` vs gui-domain load).
- **rolls one slot at a time** — the other worker keeps serving (respects the Apple 2-VMs/host cap).
- **drains each slot before recreating** — derives the slot's `workerId` from the VM MAC (`mac-<last3>`, matching `set_hostname.sh`) and waits for its current task to finish via the Taskcluster **public** API (no creds), with a timeout that *proceeds* rather than hangs. Gated on `tart.tc_worker_pool` (empty → mechanical recreate, e.g. builder hosts).

New knobs (`tart.tc_root_url` / `tc_worker_pool` / `drain_timeout`) are passed through; `tc_worker_pool` is set for the tester role.

## Scope / boundaries

Fleet-wide orchestration and TC **quarantine** (which needs creds + pool knowledge) stay one layer up — the on-network reprovision runner / Hangar trigger — which calls this primitive per host. Builder hosts (no `tc_worker_pool`) get the old mechanical behavior, just corrected for their launchd type.

Needs on-host validation (the drain + daemon cycle) before wide use; logic verified via pre-commit (puppet-validate + epp-validate + puppet-lint).

### DIFF
--- a/data/roles/tart_worker.yaml
+++ b/data/roles/tart_worker.yaml
@@ -40,6 +40,10 @@ tart:
   # and shares it into each VM via `tart run --dir`.
   inject_vault: false
   vault_role: 'gecko_t_osx_1500_m_vms'
+  # TC pool the VMs join. Enables tart-update-vms.sh's per-slot drain (waits for a
+  # slot's worker to finish its current task, via the TC public API, before
+  # recreating it). Empty elsewhere = mechanical recreate with no drain.
+  tc_worker_pool: 'releng-hardware/gecko-t-osx-1500-m-vms'
 
 # Puppet-managed autologin for the VM host (see profiles::tart): base64 of the
 # admin user's /etc/kcpassword, enabling an admin GUI session on reboot (required
--- a/modules/roles_profiles/manifests/profiles/tart.pp
+++ b/modules/roles_profiles/manifests/profiles/tart.pp
@@ -54,6 +54,14 @@ class roles_profiles::profiles::tart {
   $scep_issuer_cn = lookup('tart.scep_issuer_cn', String,  'first', 'Mozilla RelOps Bootstrap CA Intermediate CA')
   $vault_dir_base = "/Users/${user}/.tart-vault"
 
+  # Rolling image-update knobs (tart-update-vms.sh). tc_worker_pool is the TC
+  # pool the VMs join (provisioner/worker-type); when set, the update drains each
+  # slot's worker (waits for its current task to finish, via the TC *public* API
+  # — no creds) before recreating it. Empty = no drain (mechanical recreate).
+  $tc_root_url    = lookup('tart.tc_root_url',    String,  'first', 'https://firefox-ci-tc.services.mozilla.com')
+  $tc_worker_pool = lookup('tart.tc_worker_pool', String,  'first', '')
+  $drain_timeout  = lookup('tart.drain_timeout',  Integer, 'first', 3600)
+
   # Autologin the VM-host user. Apple's Virtualization Framework needs an active
   # GUI (Aqua) session for `tart run` to start a VM, and on a headless host that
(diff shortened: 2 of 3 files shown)


jwmossmoz (dismissed):

rcurranmoz:
Validated this before merging and found **two defects that would have destroyed running work**. Both fixed in 822ad999 — worth a re-look since this changes behaviour, not just style.

### 1. Drain was fail open

`worker_busy()` returned "not busy" on *every* path that wasn't a positive busy signal — JSON parse failure, curl failure, worker absent from the listing, task-status fetch failure. A transient Taskcluster blip therefore read as "idle" and the roll deleted a VM mid-task.

Demonstrated against the live queue with TC made unreachable:

```
old: worker_busy  mac-ba25cd -> NOT BUSY   (would wipe the VM mid-task)
new: worker_state mac-ba25cd -> unknown    -> treated as busy -> slot skipped
```

Same class of false-negative that already wiped a running worker via the reprovision orchestrator's `is_currently_busy`.

Replaced with `worker_state()` echoing `idle|busy|unknown`, where only a *proven* idle permits recreation. Also fixed while in there:

- **Pagination** — the worker listing is paginated; a worker on a later page looked absent, and absent is read as idle. Now follows `continuationToken`.
- **Listing lag** — a worker that has just claimed still advertises its previous resolved task for a few seconds. Now requires two consecutive idle readings 30s apart.
- **Run attribution** — matches the run's `workerId`, so a task retried on a *different* worker after ours failed no longer counts as us being busy.
- **Undeterminable workerId** — previously "skipping drain" and wiping anyway; now refuses.
- **Timeout** — was "proceeding anyway"; now skips the slot.
- **Partial rolls** — exits non-zero listing skipped slots, so the caller can't record a skipped slot as success.

Verified against the live queue: running worker → `busy`, absent → `idle`, empty id → `unknown`, TC unreachable → `unknown`.

### 2. The launchd label was always empty

```bash
local i="$1" label="com.mozilla.tartworker-${i}"
```

`local` expands all its arguments before assigning any, so `${i}` resolved to the outer (unset) `i` and `label` came out as `com.mozilla.tartworker-` — no index. Confirmed on macOS bash 3.2.57:

```
$ f() { local i="$1" label="com.mozilla.tartworker-${i}"; echo "[$label]"; }; f 2
[com.mozilla.tartworker-]
```

So `bootout` targeted a nonexistent label, `|| true` swallowed the failure, and `tart delete` then removed the VM **while its launchd job was still loaded with `KeepAlive=true`** — launchd would respawn `tart run` against a VM being deleted. Split into two `local` statements.

This one means the rolling update could not have worked as written, which fits the PR never having been exercised on a real host.

### Testing

Both rendered variants (`daemon` and `agent`) pass `bash -n` and shellcheck; the empty-`SKIPPED`-array case is safe under `set -u` on bash 3.2. Still **not** exercised end-to-end on a host — a real roll bounces VMs and kills tasks, so that wants a genuinely idle host.

markcor (approved):

PR #1295, opened 2026-07-23
Bump scriptworker-scripts revision for iscript

jcristau:


Eijebong (approved):
TIL about that script stuff. Interesting

PR #1296, opened 2026-07-23
mac signing: remove mac_v3_* roles, rename shared profile to mac_signing

hneiva:
Remove the four mac_v3_* signing roles and their dedicated artifacts: role manifests, hiera data (data/roles/mac_v3_signing_*.yaml), the Bolt inventory (inventory.d/signing.yaml), and the test/integration suites.

Drop the now-dangling inventory.d/signing.yaml entry from inventory.yaml; the missing filepath was breaking Bolt inventory loading repo-wide.

Rename the shared profile roles_profiles::profiles::mac_v3_signing to ::mac_signing (now used only by the mac_v4_signing_* roles) and update all six role includes. $worker_type='mac-v3-signing' is intentionally left unchanged as it is a live telemetry/logging label.

Rename the mac-v3-signing kitchen suite to signing-secrets and refresh its stale comments. Update the terraform-fixture example in CLAUDE.md.


hneiva:
tested on a local mac signer

bhearsum (approved):
We're in the middle of incident response for 153.0; let's avoid rolling this out until that's done.

PR #1299, opened 2026-07-27
tart: self-renewing client cert for the runtime vault fetch

rcurranmoz:
## Problem

`tart-run-vm.sh` authenticates to the vault broker on **every VM launch** using the MDM/SCEP identity in the System keychain. That cert can't sustain a runtime dependency:

- Apple's `com.apple.security.scep` payload has **no self-renewal** — the cert only refreshes if the MDM re-installs the profile, and nothing does.
- relops-bootstrap adds its SCEP provisioners with **no duration claims** (`scripts/bootstrap-step-ca.sh`), so step-ca issues them with its default **24h** lifetime.

This is harmless for the flow the cert was designed for — the bootstrap script uses it once, minutes after enrollment, to fetch `vault.yaml` — which is why it went unnoticed. But measured on 2026-07-27, **all of `macmini-m4-235..239` held SCEP certs that expired Jul 18, nine days earlier**:

```
m4-235  notAfter=Jul 18 10:50:55 2026 GMT
m4-236  notAfter=Jul 18 10:50:55 2026 GMT
m4-237  notAfter=Jul 18 10:50:55 2026 GMT
m4-238  notAfter=Jul 18 10:50:55 2026 GMT
m4-239  notAfter=Jul 18 10:50:54 2026 GMT
```

Flipping `tart.inject_vault` with `cert_source: keychain` would give each slot ~a day of working fetches and then silently dead workers.

**`step ca renew` cannot fix the SCEP cert.** It's issued `KeyIsExtractable=false`, so there is no key file to renew. A separate file-based identity is required — this is not a bug in the SCEP setup, it's the wrong credential for the job.

## What this does

**New module `macos_step_cert`** — a generic self-renewing step-ca identity: `step` CLI install, CA-trust bootstrap from a fingerprint, optional single-use-token enrollment, and a `step ca renew --daemon` LaunchDaemon. Two deliberate changes from the `reprovision_runner::step_renew` helper it's modelled on:

- Enrollment re-runs when the cert is **expired**, not only absent (`unless openssl x509 -checkend 300`). `step ca renew` needs a still-valid cert, so a host powered off longer than the lifetime can't self-heal. If the token is spent this makes puppet go red — deliberate; silent expiry is the failure being fixed.
- The `--exec` kick is optional. The runner restarts a daemon on renewal; tart re-reads the PEM per launch and needs nothing.

**`profiles::tart`** — `tart.cert_source` (`keychain` | `file`), `step_cert_*` lookups, and a `fail()` if `cert_source: file` is set without `step_cert_enabled`. Secrets are top-level keys per the hiera-`first` shadowing trap already documented in that file.

**`tart-run-vm.sh`** — PEM `--cert/--key` mode, which requires `CURL_SSL_BACKEND=openssl`. Verified on m4-235 with a throwaway self-signed pair:

| backend | result |
|---|---|
| `openssl` | handshake completes, LB returns **401** (correctly rejects untrusted client cert) |
| `securetransport` | `curl: (58) SSL: Can't find the certificate "c.pem" ... in the Keychain` |

SecureTransport can't take a separate PEM key file. Also confirms the broker is reachable and enforcing mTLS from a tart host today.

**Degradation gap closed** (prereq (b) from the 07-17 pipeline handoff). Previously every failure path still passed `--dir=vault:<dir>`, so a host with no usable cert started VMs against an **empty share** and the guest waited 2 minutes for a file that would never appear. Now: use a cached vault if one exists, otherwise start with a plain `tart run` and log to **syslog** that the slot will not register. An expired cert is reported as expired rather than as an opaque TLS error.

## Risk

**Inert as merged.** `step_cert_enabled: false` and `cert_source: keychain`, so applying this changes no behavior on any host. The `tart-run-vm.sh` degradation change is the only live difference, and it strictly improves the no-vault case.

## Not done / follow-ups

1. **CA side is the gate now** — needs a JWK provisioner on step-ca to mint enrollment tokens from, plus a token in vault per host. Ideally with duration claims and `allowRenewalAfterExpiry` set (the 24h default is what caused this, and `allowRenewalAfterExpiry` removes the powered-off-too-long cliff).
2. **Untested on a host** — there's no cert to enroll with yet. Only rendering, lint, and the curl-backend behavior have been exercised. All three template variants (`file`, `keychain`, injection-off) pass `bash -n` + shellcheck; the repo's templated-script hook only exercises one.
3. `reprovision_runner::step_renew` still duplicates this logic, left untouched on purpose — live, security-reviewed cert path; refactoring it alongside a new consumer would double the risk surface. `TODO` in the module.
4. The SCEP 24h lifetime is unchanged and stays correct for the bootstrap flow it serves.
5. `relops-bootstrap/README.md:15` claims *"routine cert rotation — SCEP renews on its own, no operator action"*. Inaccurate as implemented; different repo, not touched here.

### DIFF
--- a/data/roles/tart_worker.yaml
+++ b/data/roles/tart_worker.yaml
@@ -33,17 +33,48 @@ tart:
   manage_image: false
   launchd_type: 'daemon'
   # Host-mediated worker-vault injection (see profiles::tart). Keep FALSE until
-  # the credential-free tester image (which reads the injected vault at first
-  # boot) is built + promoted — with today's baked image the injected dir is
-  # simply ignored, so flipping this early is harmless but pointless. When true,
-  # the host fetches vault-<vault_role> from the broker (mTLS via its SCEP cert)
-  # and shares it into each VM via `tart run --dir`.
+  # the host has a RENEWING client cert (see cert_source / step_cert_* below).
+  # When true, the host fetches vault-<vault_role> from the broker over mTLS and
+  # shares it into each VM via `tart run --dir`.
+  #
+  # Prereq for flipping this: step_cert_enabled + cert_source: 'file'. Flipping
+  # it while cert_source is 'keychain' gives every slot ~24h of working vault
+  # fetches and then silent failure, because the MDM/SCEP cert expires and
+  # nothing renews it. (Measured 2026-07-27: all of m4-235..239 had SCEP certs
+  # 9 days expired.) A slot that cannot fetch now degrades to a plain `tart run`
+  # with an ERROR to syslog rather than hanging the guest on an empty share.
   inject_vault: false
   vault_role: 'gecko_t_osx_1500_m_vms'
   # TC pool the VMs join. Enables tart-update-vms.sh's per-slot drain (waits for a
   # slot's worker to finish its current task, via the TC public API, before
   # recreating it). Empty elsewhere = mechanical recreate with no drain.
   tc_worker_pool: 'releng-hardware/gecko-t-osx-1500-m-vms'
+
+  # Client identity for the broker fetch: 'keychain' (MDM/SCEP, non-renewing —
+  # legacy) or 'file' (PEM pair renewed by macos_step_cert). Use 'file' for
+  # anything that fetches at runtime; see the long note in profiles::tart.
+  cert_source: 'keychain'
+
+  # Self-renewing file-based step-ca identity (macos_step_cert). Enable per-host
+  # once it has been handed an enrollment token, then flip cert_source to 'file'
+  # and inject_vault to true.
+  step_cert_enabled: false
+  # step-ca lives in GCP and is reachable from MDC1 by IP only (its name isn't in
+  # MDC1 DNS), so pin it in /etc/hosts. TLS still validates against the name.
+  step_ca_ip: 34.61.3.27
+  step_version: '0.28.2'
+  # Where the one-time enrollment token is read from. A FILE, not a hiera/vault
+  # value: /var/root/vault.yaml is 0 bytes on these hosts so the vault layer is
+  # empty, and a file also keeps the token out of puppet's catalog entirely.
+  # Consumed and deleted on successful enrollment. Mint one with
+  # relops-bootstrap scripts/mint-tart-enrollment-token.sh and drop it here over
+  # the same channel the bootstrap PKG uses for vault.yaml.
+  step_token_file: '/var/root/step-enrollment-token'
+
+# step-ca root CA fingerprint — pins trust during `step ca bootstrap`. Non-secret,
+# same value the reprovision runner uses. TOP-LEVEL key, not nested under `tart:`
+# (see the kcpassword note below for why).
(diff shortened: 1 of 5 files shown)


rcurranmoz:
Reworked and rebased onto master. Three things changed since the original.

## 1. Rebased (was `CONFLICTING`)

#1294/#1300/#1301 all landed in `tart.pp` and `tart-run-vm.sh.epp` since this was opened. Conflicts were additive — master's drain knobs and this branch's cert knobs occupy the same region — so both are kept. Now a single commit against current master.

## 2. `allowRenewalAfterExpiry` is resolved, the safe way

This was flagged as the one item wanting explicit sign-off. Asked which way a security review would land, the answer was the opposite of what I'd set, so it's reversed in relops-bootstrap#52 and already applied to the live CA:

```
jwk-tart-gecko-t-osx-1500-m-vms          allowRenewalAfterExpiry=False  dur=720h0m0s
jwk-tart-gecko-t-osx-1500-m-vms-staging  allowRenewalAfterExpiry=False  dur=720h0m0s
```

Renewal is authenticated **by the cert itself**, so permitting it after expiry makes a leaked cert renewable indefinitely — an unbounded bearer credential, which defeats the point of short-lived certs. "Revoke at the CA" depends on noticing the theft, and nothing monitors these hosts.

The operational reason I'd set it is handled instead by **720h certs** (renewing at ~20 days, covering realistic downtime) plus re-enrollment past that window — which is why enrollment triggers on an **expired** cert, not just a missing one.

## 3. The enrollment token no longer comes from vault — it couldn't

The original took it as a vault-delivered `Sensitive` value. That cannot work: **`/var/root/vault.yaml` is 0 bytes on every tart host**, so the vault hiera layer is empty and anything routed through it resolves to `undef`. (Same reason `tart_autologin_kcpassword` and #1284 have never taken effect — worth knowing independently of this PR.)

Now `tart.step_token_file`, default `/var/root/step-enrollment-token`:

- The token **never enters puppet's catalog or logs** — the exec reads it in-shell
- It matches a delivery path that already exists: the bootstrap PKG / reprovision orchestrator already drops `/var/root/vault.yaml` over the same channel
- Deleted only after a **successful** enrollment (`set -e` before the `rm`), so a failed attempt retries rather than silently burning the token
- Absent token + no valid cert is a no-op, not a puppet failure — that's the normal state of a host not yet enrolled

Rendered command, verified as valid bash:

```
/bin/bash -c 'set -e; tok=$(/bin/cat /var/root/step-enrollment-token); \
  /usr/local/bin/step ca certificate macmini-m4-235 \
  /etc/step-cert/tart-client.crt /etc/step-cert/tart-client.key \
  --token "$tok" --force; /bin/rm -f /var/root/step-enrollment-token'
```

## Unchanged

Still inert as merged — `step_cert_enabled: false`, `cert_source: keychain`. The only live behaviour change is `tart-run-vm.sh` degrading to a plain `tart run` instead of passing an empty vault share, which strictly improves the no-cert case.

Still verified end-to-end on m4-235: correct SPIFFE SAN, broker fetch HTTP 200/1440 bytes with the PEM identity, `step ca renew` rotating the serial while preserving the SAN.

pre-commit clean (puppet-lint, EPP validate, templated-script syntax check).

PR #1305, opened 2026-07-30
chore: remove amadan from signing users

ahal:


jcristau:
Does this also clean things up, or not until/unless we reimage?

hneiva:
> Does this also clean things up, or not until/unless we reimage?

No, for that we'd need to add it to https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/modules/macos_people_remover/files/user_cleanup.sh#L3

jcristau (approved):

